PR #13, opened 2017-03-07
Selenium refactoring

jeddcuison:
Changes for this pull request is generally  to refactor the existing selenium tests. The objective is to establish a better way of coding xpath expressions, optimizing test execution, introduction of test suite to group tests that have dependencies with one another and to make selenium UI testing more robust and reliable.

While working on this, some of the tests exposed problems in the UI that is why one would notice 2 UI commits which is about

- stomp 
- loading icon

Those commits where cherry picked from the old repository which is the reason why the commit dates are old. In addition one would notice that the selenium commit proper is just one commit. It's also because those changes were from an the old repository but instead of cherry picking each and everyone of the commits, I just opted to copy the codes since they all reside in the selenium module.


jwala-integration-test/build.gradle
```diff
@@ -12,7 +12,6 @@ dependencies {
 }
 
 test {
-    systemProperties = [
-            "selenium.property.file": System.getProperty("selenium.property.file")
-    ]
+    systemProperty "selenium.property.path", System.getProperty("selenium.property.path")
+    exclude '**/testSuiteClasses/**'
```
akinny: Shouldn't this be  exclude '**/testSuite/**'?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/testSuite/JwalaTest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.openqa.selenium.support.ui.WebDriverWait;
+
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Wraps Jwala specific UI actions e.g. click tab, etc also contains methods used commonly in tests
+ * Created by Jedd Cuison on 2/28/2017
+ */
+public class JwalaTest {
```
jmahmoodcc: Since we're hard coding the Chrome Web Driver in this class we should modify the name of the class to indicate that it currently only works with Chrome

jmahmoodcc: Also this appears to be a utility class, so I would remove Test from the name as well

jmahmoodcc: Oh nevermind, I see that your extending this class for the other tests now

jmahmoodcc (changes requested):
Before merging I'm requesting that the external properties tests and the history pop-up tests be recreated with the new testing structure. These were originally created because of failing regression tests that no one noticed for a long time, so we wanted to automate those tests. I like all the changes we made, but we still need to honor the gaps we had already closed with the initial implementation.

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/testSuite/JwalaTest.java
jeddcuison: Good catch! I'll make it generic. I don't want to make JwalaTest specific to Chrome since it is inherited by all test suite classes.

jeddcuison: DONE

jwala-integration-test/build.gradle
jeddcuison: Done renamed to testsuite

jeddcuison:
This was done under a fork and since I deleted the fork this pull request won't be able to merge to master. I'm creating another pull request.

PR #20, opened 2017-03-11
Linux: query jvm state

jmahmoodcc:
When the reverse heartbeat kicks in, one of the final checks is an 'sc query' of the JVM. This is a windows specific command and will not work on Linux.

These changes do the following:
* Move the hard coded query command from java to a bash script
* The script checks for the OS (Windows or Linux) and then runs the appropriate query command
* The script is copied to the JVM's bin directory as part of JVM generation

KNOWN ISSUE
* In the Linux script jvm-service.sh, the status method is not returning correctly when a JVM is stopped. Instead of returning "jvm-name is stopped" it's returning "(pid ) is running...' where the PID is blank because the service is stopped.


jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/spring/component/JvmStateResolverWorker.java
```diff
@@ -34,7 +34,8 @@
 public class JvmStateResolverWorker {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(JvmStateResolverWorker.class);
-    private static final String STOPPED = "STOPPED";
+    private static final String WINDOWS_SVC_STOPPED = "STOPPED";
+    private static final String LINUX_SVC_STOPPED = "is stopped";
```
akinny: We control the status string in our Linux shell scripts(jvm-service.sh), let's use "STOPPED" so we don't have to make this distinction.

akinny (commented):
can you make the script change?

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/spring/component/JvmStateResolverWorker.java
jmahmoodcc: Ok, I'll make that change

jmahmoodcc:
I'll make the change to the jvm-service.sh script as well

akinny (approved):

PR #25, opened 2017-03-14
Copy the JVM service scripts

jmahmoodcc:
Due to directory changes between installed versions and the latest release, the JVM service scripts are being moved to a common location in the .jwala directory of the SSH user
* Move the start-service, stop-service, thread-dump, heap-dump, and service-status scripts to .jwala
* Before executing the script check to see if it exists, copy it to the directory if not and make the scripts in the directory executable


jwala-services/src/main/java/com/cerner/jwala/control/jvm/command/JvmCommandFactory.java
```diff
@@ -46,56 +44,105 @@
...
+            copyScriptToRemoteDestination(jvm, scriptName, destAbsolutePath);
+        } else {
+            LOGGER.info("{} already exists. Continue with script execution", scriptName);
+        }
+    }
+
+    private void copyScriptToRemoteDestination(Jvm jvm, String scriptName, String destAbsolutePath) {
+        LOGGER.info("{} does not exist at remote location. Performing secure copy.", scriptName);
+
+        final String destDir = new File(getFullPathScript(jvm, scriptName)).getParent();
```
akinny: This will not work cross OS generate the path using vars.properties, jvm name and script name. Don't use new File()

jmahmoodcc: I've tested this on a Redhat VM and it is working. I can try it out on our Linux CI loop later, too but I'm fairly certain it will work there as well.

Which part of the new File() doesn't work in Linux? Java is supposed to be "write once, run anywhere", why would we write different code to handle file objects differently based on the OS?

akinny: You are right, this will work on any OS, the issue will be it generates the OS specific paths. So jwala on Linux(or Mac) will not be able to Windows or vice versa

jmahmoodcc: Ok, that's definitely a problem. How did we get around this previously?

PR #28, opened 2017-03-16
tomcat manager.xml

jmahmoodcc:
Place a manager.xml in the SSL directory of the JVM to allow monitoring of SSL enabled applications running on the JVM.

To enable the manager app in SSL 3 conditions need to be met:

1. In vars.properties the tomcat.manager.xml.ssl.path property must be set
2. In the data/templates directory of jwala there must be a manger.xml template
3. And for the Operations tab in the UI to use the SSL URL for the manager app, in vars.properties the tomcat.manager.protocol property must be set

The manager.xml is added to the generated JVM jar as an internal operation and does not show up as a resource in the UI.

### DIFF
--- a/jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/ManagedJvmBuilder.java
+++ b/jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/ManagedJvmBuilder.java
@@ -26,6 +26,7 @@ import java.nio.file.*;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.*;
 
+import static com.cerner.jwala.common.properties.PropertyKeys.PATHS_RESOURCE_TEMPLATES;
 import static com.cerner.jwala.common.properties.PropertyKeys.TOMCAT_MANAGER_XML_SSL_PATH;
 
 /**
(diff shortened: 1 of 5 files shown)


akinny:
Where is the manager.xml template?

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/ManagedJvmBuilder.java
```diff
@@ -75,11 +77,54 @@ public ManagedJvmBuilder build() {
...
+        } catch (IOException e) {
+            LOGGER.error("Error adding the manager.xml to the generated staging directory. Could not write content {} to {}", generatedManagerXmlContent, generatedDestDir, e);
+            throw new JvmServiceException(e);
+        }
+        return this;
+    }
+
+    private String generateManagerXml(File srcManagerXmlFile) {
+        try {
+            final FileInputStream managerXmlTemplateContent = new FileInputStream(srcManagerXmlFile);
```
akinny, line 113: close stream in finally block

jmahmoodcc:
The manager.xml is applied in our chef cookbook. This code allows anyone to setup the manager app through SSL in their own jwala deployment using the properties and providing their own manager.xml template, but it is not provided as a default configuration.

PR #34, opened 2017-03-28
Fix bug wherein a newly defined web server's status is set to STOPPED

jeddcuison:
When the user adds a web server, the state of the web server should be NEW until the user deploys it. The service that resolves the state of the web servers should not process web servers with NEW state.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-buildVersion=0.0.185
+buildVersion=0.0.186-SNAPSHOT
 
 #daemon should be false for CI builds
 org.gradle.daemon=false


jwala-services/src/main/java/com/cerner/jwala/service/webserver/component/WebServerStateSetterWorker.java
```diff
@@ -113,37 +122,18 @@ public void pingWebServer(final WebServer webServer) {
...
-        if (isWebServerBusy(webServer)) {
-            LOGGER.debug("Cannot ping web server {} since it is busy. Details: {}", webServer.getName(), webServer);
+        if (WebServerReachableState.WS_NEW.equals(webServer.getState())) {
+            LOGGER.debug("Cannot ping web server {} with a status of \"NEW\"", webServer.getName());
             return false;
         }
 
-        if (inMemoryStateManagerService.get(webServer.getId()) == WebServerReachableState.WS_NEW) {
-            LOGGER.debug("Cannot ping web server {} with a status of \"NEW\"", webServer.getName(), webServer);
+        if (isWebServerBusy(webServer)) {
```
jmahmoodcc: Can we rename this to isWebServerStartingOrStopping() or something similar since that's what is being checked?
When I see isWebServerBusy() I think it's processing something critical like a client request and shouldn't be pinged.

jwala-services/src/main/java/com/cerner/jwala/service/webserver/component/WebServerStateSetterWorker.java
```diff
@@ -113,37 +122,18 @@ public void pingWebServer(final WebServer webServer) {
     }
 
     /**
-     * Checks if the web server is in webServerToPingSet, if it's not in the set it will be added to the set.
-     * This method is thread safe.
-     * @param webServer the web server
-     * @return true if the web server is in webServerToPingSet
-     */
-    private boolean isInWebServerToPingSet(final WebServer webServer) {
```
jmahmoodcc: Why remove this method and extract it to pingWebServer? It seems like we could still use it and make the pingWebServer method easier to read

jwala-services/src/main/java/com/cerner/jwala/service/webserver/component/WebServerStateSetterWorker.java
```diff
@@ -113,37 +122,18 @@ public void pingWebServer(final WebServer webServer) {
...
-            LOGGER.debug("Cannot ping web server {} since it is busy. Details: {}", webServer.getName(), webServer);
+        if (WebServerReachableState.WS_NEW.equals(webServer.getState())) {
+            LOGGER.debug("Cannot ping web server {} with a status of \"NEW\"", webServer.getName());
             return false;
         }
 
-        if (inMemoryStateManagerService.get(webServer.getId()) == WebServerReachableState.WS_NEW) {
-            LOGGER.debug("Cannot ping web server {} with a status of \"NEW\"", webServer.getName(), webServer);
+        if (isWebServerBusy(webServer)) {
+            LOGGER.debug("Cannot ping web server {} since it is busy. Details: {}", webServer.getName(), webServer);
```
jmahmoodcc: instead of logging the entire web server object I would just log the web server state since that's what isWebServerBusy() is checking

jmahmoodcc (commented):
Looks ok. There are some minor changes I would make, but look at my comments and let me know.

jwala-services/src/main/java/com/cerner/jwala/service/webserver/component/WebServerStateSetterWorker.java
jeddcuison: The problem is this method adds web server to the set which is not apparent by just looking at it. I was actually the one who wrote this and yet I got confused when I came back to do work on it. I can't really separate the "add" part since it should be under a synchronized clause.  

In addition it is an anti pattern since it does something else rather than check if the web server is in the set. I can rename the method but it seems awkward e.g. isInWebServerToPingSetAddIfItIsNot :)

Maybe we can do something this ?

   webServerPingValidator(webServer).canBePinged().ifYes().isBeingPinged().ifNot().addToPingQeue()

Then the validator can be synchronized

akinny (approved):

PR #36, opened 2017-03-29
jvm and webserver validation for service name conflict

rahulsayini:


jwala-services/src/main/java/com/cerner/jwala/service/configuration/service/AemServiceConfiguration.java
```diff
@@ -97,23 +97,23 @@
 @EnableAsync
 @EnableScheduling
 @ComponentScan({"com.cerner.jwala.service.webserver.component",
-                "com.cerner.jwala.service.state",
```
jeddcuison: The alignment was fine already, I don't think you needed to change it.

jwala-services/src/main/java/com/cerner/jwala/service/configuration/service/AemServiceConfiguration.java
```diff
@@ -170,12 +170,12 @@ public JvmService getJvmService(final GroupPersistenceService groupPersistenceSe
...
                                     @Value("${spring.messaging.topic.serverStates:/topic/server-states}") final String topicServerStates,
                                     final JvmControlService jvmControlService, final HistoryFacadeService historyFacadeService,
-                                    final FileUtility fileUtility) {
+                                    final FileUtility fileUtility, final WebServerPersistenceService webServerPersistenceService) {
         final JvmPersistenceService jvmPersistenceService = aemPersistenceServiceConfiguration.getJvmPersistenceService();
         return new JvmServiceImpl(jvmPersistenceService, groupPersistenceService, applicationService,
                 messagingTemplate, groupStateNotificationService, resourceService,
                 clientFactoryHelper, topicServerStates, jvmControlService, binaryDistributionService, binaryDistributionLockManager,
-                historyFacadeService, fileUtility);
+                historyFacadeService, webServerPersistenceService,fileUtility);
```
jeddcuison: Try to be consistent with spaces, add a space after webServerPersistenceService

jwala-services/src/main/java/com/cerner/jwala/service/configuration/service/AemServiceConfiguration.java
```diff
@@ -191,21 +191,23 @@ public BalancerManagerService getBalancermanagerService(final GroupService group
...
+                new BalancerManagerHttpClient(), historyFacadeService);
     }
 
     @Bean(name = "webServerService")
     public WebServerService getWebServerService(final ResourceService resourceService,
                                                 @Qualifier("webServerInMemoryStateManagerService")
                                                 final InMemoryStateManagerService<Identifier<WebServer>, WebServerReachableState> inMemoryStateManagerService,
-                                                @Value("${paths.resource-templates:../data/templates}") final String templatePath) {
+                                                @Value("${paths.resource-templates:../data/templates}") final String templatePath,
+                                                final JvmPersistenceService jvmPersistenceService) {
```
jeddcuison: Although I prefer constructor injection, I remember the team decided to do class variable auto-wiring instead to keep the constructor from having too many parameters (which is a good point) 

    class WebServerServiceImpl 
       
        @Autowired
        private JvmPersistenceService jvmPersistenceService;

        ....

    }

However, please confirm with the others if you can do constructor injection for now since we're in a tight deadline

jwala-services/src/main/java/com/cerner/jwala/service/configuration/service/AemServiceConfiguration.java
```diff
@@ -232,8 +234,8 @@ public ApplicationService getApplicationService(final JvmPersistenceService jvmP
 
     @Bean(name = "jvmControlService")
     public JvmControlService getJvmControlService(
-                                                  final JvmStateService jvmStateService,
-                                                  final HistoryFacadeService historyFacadeService) {
+            final JvmStateService jvmStateService,
```
jeddcuison: Since you modified the formatting already, wouldn't it be better if you formatted it this way ?

    public JvmControlService getJvmControlService(final JvmStateService jvmStateService, 
                                                  final HistoryFacadeService historyFacadeService) {
        ....
    }

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -115,11 +120,18 @@ public JvmServiceImpl(final JvmPersistenceService jvmPersistenceService,
...
         this.historyFacadeService = historyFacadeService;
+        this.webServerPersistenceService = webServerPersistenceService;
         this.fileUtility = fileUtility;
     }
 
 
     protected Jvm createJvm(final CreateJvmRequest aCreateJvmRequest) {
+        List<WebServer> webServerList = webServerPersistenceService.getWebServers();
+        for (WebServer webserver : webServerList) {
+            if (aCreateJvmRequest.getJvmName().equals(webserver.getName())) {
```
jeddcuison: Use equalsIgnoreCase since we don't want the names to be case sensitive.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -115,11 +120,18 @@ public JvmServiceImpl(final JvmPersistenceService jvmPersistenceService,
         this.binaryDistributionService = binaryDistributionService;
         this.binaryDistributionLockManager = binaryDistributionLockManager;
         this.historyFacadeService = historyFacadeService;
+        this.webServerPersistenceService = webServerPersistenceService;
         this.fileUtility = fileUtility;
     }
 
 
     protected Jvm createJvm(final CreateJvmRequest aCreateJvmRequest) {
```
jeddcuison: I can't help notice that we have a createJvm method in this service. The purpose of the JvmStateService is to provide state related functionality. Can you ask @akinny why it's here ? I would prefer it to be in JvmService.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -115,11 +120,18 @@ public JvmServiceImpl(final JvmPersistenceService jvmPersistenceService,
...
+        this.webServerPersistenceService = webServerPersistenceService;
         this.fileUtility = fileUtility;
     }
 
 
     protected Jvm createJvm(final CreateJvmRequest aCreateJvmRequest) {
+        List<WebServer> webServerList = webServerPersistenceService.getWebServers();
+        for (WebServer webserver : webServerList) {
+            if (aCreateJvmRequest.getJvmName().equals(webserver.getName())) {
+                throw new EntityExistsException("Webserver already exists with this name "+ aCreateJvmRequest.getJvmName());
```
jeddcuison: In addition I'm not in favor of using EntityExistException because at first glance of an exception, it doesn't tell where the exception occurred. I would prefer that you use JvmStateServiceException.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -143,7 +155,6 @@ protected Jvm createAndAssignJvm(final CreateJvmAndAddToGroupsRequest aCreateAnd
     public Jvm createJvm(CreateJvmAndAddToGroupsRequest createJvmAndAddToGroupsRequest, User user) {
```
jeddcuison: Again ask @akinny if this can be transferred to JvmService.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -246,7 +257,12 @@ public Jvm getJvm(final String jvmName) {
     public Jvm updateJvm(final UpdateJvmRequest updateJvmRequest, final boolean updateJvmPassword) {
 
         updateJvmRequest.validate();
-
+        List<WebServer> webServerList = webServerPersistenceService.getWebServers();
+        for (WebServer webserver : webServerList) {
+            if (updateJvmRequest.getNewJvmName().equals(webserver.getName())) {
```
jeddcuison: Use equalsIgnoreCase since we don't want the names to be case sensitive

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -622,7 +638,15 @@ private void secureCopyJvmConfigJar(Jvm jvm, String jvmConfigJar, User user) thr
 
     private void deployJvmConfigJar(Jvm jvm, User user, String jvmJar) throws CommandFailureException {
```
jeddcuison: My bad I missed some of the code reviews. I don't think this should be in a state service, please ask @akinny if we can transfer this

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -622,7 +638,15 @@ private void secureCopyJvmConfigJar(Jvm jvm, String jvmConfigJar, User user) thr
 
     private void deployJvmConfigJar(Jvm jvm, User user, String jvmJar) throws CommandFailureException {
         final String parentDir = ApplicationProperties.get("remote.paths.instances");
-        CommandOutput execData = jvmControlService.controlJvm(
+        CommandOutput execData = jvmControlService.executeCreateDirectoryCommand(jvm, parentDir+"/"+jvm.getJvmName());
```
jeddcuison: Be consistent with spacing

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -622,7 +638,15 @@ private void secureCopyJvmConfigJar(Jvm jvm, String jvmConfigJar, User user) thr
...
     private void deployJvmConfigJar(Jvm jvm, User user, String jvmJar) throws CommandFailureException {
         final String parentDir = ApplicationProperties.get("remote.paths.instances");
-        CommandOutput execData = jvmControlService.controlJvm(
+        CommandOutput execData = jvmControlService.executeCreateDirectoryCommand(jvm, parentDir+"/"+jvm.getJvmName());
+        if (execData.getReturnCode().wasSuccessful()) {
+            LOGGER.info("Successfully created the parent directory {}", parentDir);
+        } else {
+            String standardError = execData.getStandardError().isEmpty() ? execData.getStandardOutput() : execData.getStandardError();
+            LOGGER.error("Deploy command completed with error trying to extract and back up JVM config {} :: ERROR: {}", jvm.getJvmName(), standardError);
+            throw new InternalErrorException(FaultType.REMOTE_COMMAND_FAILURE, standardError.isEmpty() ? CommandOutputReturnCode.fromReturnCode(execData.getReturnCode().getReturnCode()).getDesc() : standardError);
```
jeddcuison: I know InternalErrorException has been used in this application for some time now, I'm not in favor of it, my 2 cents is to throw an service exception that relates to this class. The reason is when you see the exception, you would immediately see InternalErrorException - it does not tell you right away where the problem happened.

jwala-services/src/main/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImpl.java
```diff
@@ -50,17 +53,21 @@
...
 
+    private final JvmPersistenceService jvmPersistenceService;
+
     public WebServerServiceImpl(final WebServerPersistenceService webServerPersistenceService,
                                 final ResourceService resourceService,
                                 @Qualifier("webServerInMemoryStateManagerService")
                                 final InMemoryStateManagerService<Identifier<WebServer>, WebServerReachableState> inMemoryStateManagerService,
                                 final String templatePath,
-                                final BinaryDistributionLockManager binaryDistributionLockManager) {
+                                final BinaryDistributionLockManager binaryDistributionLockManager,
```
jeddcuison: Ask the team if you should use class variable auto-wiring since that is what we talked about doing going forward

jwala-services/src/main/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImpl.java
```diff
@@ -76,6 +83,13 @@ public WebServer createWebServer(final CreateWebServerRequest createWebServerReq
                                      final User aCreatingUser) {
         createWebServerRequest.validate();
 
+        List<Jvm> jvmList = jvmPersistenceService.getJvms();
+        for (Jvm jvm : jvmList) {
+            if(createWebServerRequest.getName().equals(jvm.getJvmName())){
```
jeddcuison: Use equalsIgnoreCase

jwala-services/src/main/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImpl.java
```diff
@@ -134,6 +148,12 @@ public WebServer updateWebServer(final UpdateWebServerRequest anUpdateWebServerC
                                      final User anUpdatingUser) {
         anUpdateWebServerCommand.validate();
 
+        List<Jvm> jvmList = jvmPersistenceService.getJvms();
+        for (Jvm jvm : jvmList) {
+            if(anUpdateWebServerCommand.getNewName().equals(jvm.getJvmName())){
```
jeddcuison: Use equalsIgnoreCase

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -246,7 +257,12 @@ public Jvm getJvm(final String jvmName) {
     public Jvm updateJvm(final UpdateJvmRequest updateJvmRequest, final boolean updateJvmPassword) {
 
         updateJvmRequest.validate();
-
+        List<WebServer> webServerList = webServerPersistenceService.getWebServers();
+        for (WebServer webserver : webServerList) {
+            if (updateJvmRequest.getNewJvmName().equals(webserver.getName())) {
+                throw new EntityExistsException("Webserver already exists with this name "+ updateJvmRequest.getNewJvmName());
```
jeddcuison: Throw JvmServiceException

jeddcuison (changes requested):
- Use equalsIgnoreCase since names should not be case sensitive
- Write unit tests for these fixes

PeterTeunissen:
Just my $0.02: I don't understand why people format their code manually.... Control-Shift-F is your friend.... (oh... and by the way: so is Control-Shift-O). You can hook it all up to Control-S while you save the file....

jwala-services/src/main/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImpl.java
```diff
@@ -76,6 +83,14 @@ public WebServer createWebServer(final CreateWebServerRequest createWebServerReq
                                      final User aCreatingUser) {
         createWebServerRequest.validate();
 
+        List<Jvm> jvmList = jvmPersistenceService.getJvms();
+        for (Jvm jvm : jvmList) {
```
PeterTeunissen: Same here: why not call findJvmByExactName method here?

PeterTeunissen (changes requested):
See comments: instead of getting lists and looping, why not try to get by name?

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
```diff
@@ -1006,7 +1009,7 @@ public JvmService getJvmService() {
             return new JvmServiceImpl(mockJvmPersistenceService, mockGroupPersistenceService, mockApplicationService,
                     mockMessagingTemplate, mockGroupStateNotificationService, mockResourceService, mockClientFactoryHelper,
                     "/topic/server-states", mockJvmControlService, mockBinaryDistributionService, mockBinaryDistributionLockManager,
-                    mockHistoryFacadeService, new FileUtility());
+                    mockHistoryFacadeService, mockWebServerPersistenceService, new FileUtility());
         }
 
     }
```
PeterTeunissen: You made this class compile, but where is the junit test that actually tests that trying to add a dupe will throw an error?

jwala-services/src/test/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImplTest.java
```diff
@@ -132,7 +136,7 @@ public void setUp() throws IOException {
         mockWebServers11.add(mockWebServer);
         mockWebServers12.add(mockWebServer2);
 
-        wsService = new WebServerServiceImpl(webServerPersistenceService, resourceService, inMemService, StringUtils.EMPTY, binaryDistributionLockManager);
+        wsService = new WebServerServiceImpl(webServerPersistenceService, resourceService, inMemService, StringUtils.EMPTY, binaryDistributionLockManager, jvmPersistenceService);
```
PeterTeunissen: You made this class compile, but where is the junit test that actually tests that trying to add a dupe will throw an error?

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
jmahmoodcc: Was this code block added as part of a merge?
I don't see how this pertains to solving the problem, and I consider it to be a risky change at this point in the release.

jwala-services/src/main/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImpl.java
jmahmoodcc: I would prefer all new construction injection parameters be auto wired. But like Jedd said earlier, this might require a lot of re-work in the unit test.

jwala-services/src/main/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImpl.java
```diff
@@ -76,6 +83,13 @@ public WebServer createWebServer(final CreateWebServerRequest createWebServerReq
...
+            if(createWebServerRequest.getName().equals(jvm.getJvmName())){
+                throw new EntityExistsException("Jvm already exists with this name "+ createWebServerRequest.getName());
```
jmahmoodcc: Like Jedd stated earlier, throw a WebServerServiceException here. I'm guilty of this as well, but now I know what to look for in the future.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -126,13 +125,12 @@ public JvmServiceImpl(final JvmPersistenceService jvmPersistenceService,
...
-        List<WebServer> webServerList = webServerPersistenceService.getWebServers();
-        for (WebServer webserver : webServerList) {
-            if (aCreateJvmRequest.getJvmName().equalsIgnoreCase(webserver.getName())) {
-                LOGGER.error("Webserver already exists with this name {}", aCreateJvmRequest.getJvmName());
-                throw new EntityExistsException("Webserver already exists with this name "+ aCreateJvmRequest.getJvmName());
-            }
+
+        if (null != webServerPersistenceService.findWebServerByName(aCreateJvmRequest.getJvmName())) {
+            LOGGER.error("Webserver already exists with this name {}", aCreateJvmRequest.getJvmName());
+            throw new EntityExistsException("Webserver already exists with this name "+ aCreateJvmRequest.getJvmName());
```
jmahmoodcc: Change all the EntityExistsException to the name of the class exceptions

jmahmoodcc (changes requested):
I think you got most of the changes requested from Peter and Jedd. It looks like the only thing left are to throw different exceptions based on the class their in.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
akinny: I deleted the following as part of my fix, can you double check with master? This will cause the creation of additional empty jvm  backup folder, I think you need to remove this.
jvmControlService.executeCreateDirectoryCommand(jvm, parentDir+"/"+jvm.getJvmName());

akinny (changes requested):
Make sure to check my comment

PeterTeunissen (approved):

PR #44, opened 2017-03-31
check for existing associations before removing media

rahulsayini:


jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -85,8 +90,20 @@ public JpaMedia create(final Map<String, String> mediaDataMap, final Map<String,
...
-        repositoryService.delete(jpaMedia.getLocalPath().getFileName().toString());
+        checkForExistingAssociationsBeforeRemove(name);
         mediaDao.remove(jpaMedia);
+        repositoryService.delete(jpaMedia.getLocalPath().getFileName().toString());
+    }
+
+    private void checkForExistingAssociationsBeforeRemove(String name) {
+        List<Jvm> jvmList = jvmPersistenceService.getJvms();
+        for (Jvm jvm : jvmList) {
+            if (jvm.getJdkMedia().getName().equals(name)) {
```
PeterTeunissen: Should this check be changed to ignore case?

jwala-webapp/src/main/webapp/resources/js/jwala/v1/service/mediaService.js
```diff
@@ -20,7 +20,7 @@ var mediaService = {
         return serviceFoundation.promisedPut("v1.0/media", "json", JSON.stringify(jsonData));
     },
     deleteMedia: function(name, caughtCallback) {
-        return serviceFoundation.promisedDel("v1.0/media/" + encodeURIComponent(name), "json");
+        return serviceFoundation.del("v1.0/media/" + encodeURIComponent(name), "json", caughtCallback);
```
PeterTeunissen: I'm not a UI expert at all..... but you changed from .promisedDel to .del     is that OK? Check with Jed/Jeff ?

PeterTeunissen (commented):
Leave a comment

jwala-webapp/src/main/webapp/resources/js/jwala/v1/service/mediaService.js
jeddcuison: Peter is correct going forward we should be using promisedDel since it is more straight forward and avoids "callback hell" - http://callbackhell.com/. We will be removing the regular methods once it is confirmed that we're not using them anymore.

In addition, the UI might not work properly with regards to error handling since the code using this service is written to use promises caught

    confirmDeleteCallback: function() {
        var self = this;
        ServiceFactory.getMediaService().deleteMedia(this.state.selectedMedia["str-name"]).then(function(response){
            self.refs.confirmDeleteMediaDlg.close();
            self.loadTableData(function(){
                self.state.selectedMedia = null;
            });
        }).caught(function(response){
            $.errorAlert(response);
        });
    }

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -85,8 +90,20 @@ public JpaMedia create(final Map<String, String> mediaDataMap, final Map<String,
...
         mediaDao.remove(jpaMedia);
+        repositoryService.delete(jpaMedia.getLocalPath().getFileName().toString());
+    }
+
+    private void checkForExistingAssociationsBeforeRemove(String name) {
+        List<Jvm> jvmList = jvmPersistenceService.getJvms();
+        for (Jvm jvm : jvmList) {
+            if (jvm.getJdkMedia().getName().equalsIgnoreCase(name)) {
+                LOGGER.error("The media {} cannot be deleted because it is still configured with Jvm", name);
+                throw new MediaServiceException(MessageFormat
```
jeddcuison: Please avoid redundancy, this can be rewritten as

    final String msg = MessageFormat.format("The media {0} cannot be deleted because it is associated with a JVM or JVMs");
    LOGGER.error(msg);
    throw new MediaServiceException(msg);

Also please take note I changed the message a bit to make it more concise

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -85,8 +90,20 @@ public JpaMedia create(final Map<String, String> mediaDataMap, final Map<String,
     @Transactional
     public void remove(final String name) {
         final JpaMedia jpaMedia = mediaDao.find(name);
-        repositoryService.delete(jpaMedia.getLocalPath().getFileName().toString());
+        checkForExistingAssociationsBeforeRemove(name);
         mediaDao.remove(jpaMedia);
+        repositoryService.delete(jpaMedia.getLocalPath().getFileName().toString());
+    }
+
+    private void checkForExistingAssociationsBeforeRemove(String name) {
```
jeddcuison: Since this method just checks for JVM association, just name it as checkForJvmAssociation then while you're at it can you put java doc so that it's easy for people to see that it throws an exception when the check fails.

jeddcuison (changes requested):
Revert back to promisedDel for deleteMedia

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -85,8 +90,20 @@ public JpaMedia create(final Map<String, String> mediaDataMap, final Map<String,
...
+        List<Jvm> jvmList = jvmPersistenceService.getJvms();
+        for (Jvm jvm : jvmList) {
+            if (jvm.getJdkMedia().getName().equalsIgnoreCase(name)) {
```
akinny: Since we allow null in jdkmedia, it might throw NPE. change it to 
name.equalsIgnoreCase(jvm.getJdkMedia().getName())

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -85,8 +90,20 @@ public JpaMedia create(final Map<String, String> mediaDataMap, final Map<String,
     @Transactional
     public void remove(final String name) {
         final JpaMedia jpaMedia = mediaDao.find(name);
-        repositoryService.delete(jpaMedia.getLocalPath().getFileName().toString());
+        checkForExistingAssociationsBeforeRemove(name);
```
akinny: I would make checkForExistingAssociationsBeforeRemove(name) return a boolean and throw an exception if true[may be rename to associationExists(name)], else remove it. This way we can reuse this method

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
rahulsayini: done

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
rahulsayini: done

jwala-webapp/src/main/webapp/resources/js/jwala/v1/service/mediaService.js
rahulsayini: done

PR #55, opened 2017-04-04
Only set web server state to unreachable once after a state change

jeddcuison:


jwala-services/src/main/java/com/cerner/jwala/service/webserver/component/WebServerStateSetterWorker.java
```diff
@@ -102,15 +102,17 @@ public void pingWebServer(final WebServer webServer) {
             response = request.execute();
             LOGGER.debug("Web server {} status code = {}", webServer.getName(), response.getStatusCode());
 
-            if (response.getStatusCode() == HttpStatus.OK) {
+            if (HttpStatus.OK.equals(response.getStatusCode())) {
```
akinny, line 105: HttpStatus.OK is enum OK(200, "OK"), will this be true or should we look for OK.value(), can you double check?

jeddcuison: response.getStatusCode returns the enum HttpStatus that is why equals is fine. I also tested the code before committing. Good comment though, we can never be too careful with web server status.

PR #63, opened 2017-04-06
Add group state information saving and retrieval to and from a map

jeddcuison:
This change addresses the problem wherein getting the JVM and web server start stop counts results to a null content when done in a loop with a small time interval - say 1 second. 

The approach takes advantage of the fact that the group notification service updates the JVM and web server counts when there is a state change. So to prevent the REST api from going to the database just to get the JVM and web server counts, the data that the group notification service fetches is saved in a map. That is the data that is presented to the REST API when a request comes in. It's the same data that is used by the UI so it should be accurate.


jwala-services/src/main/java/com/cerner/jwala/service/group/impl/spring/component/GroupStateNotificationServiceImpl.java
```diff
@@ -66,22 +76,51 @@ public void retrieveStateAndSend(final Identifier id, final Class aClass) {
...
+            final CurrentState<Group, OperationalState> groupState = new CurrentState<>(new Identifier<>(group.getId()),
+                    null, DateTime.now(), StateType.GROUP, webServerCount, webServerStartedCount,
+                    webServerStoppedCount, jvmCount, jvmStartedCount, jvmStoppedCount, jvmForciblyStoppedCount);
+
+            if (send) {
                 messagingService.send(groupState);
-                LOGGER.debug("Group '{}' state = {}", group.getName(), groupState);
+            }
+
+            groupStateMap.put(group.getName(), groupState);
```
akinny: make this synchronized block

jeddcuison: Hi Arvindo, fetch state is used inside a synchronized block already. With regards to someone reading the map while writing I know there will be a race condition but I'm not overly concerned with the reading part due to performance concerns. In addition the said operation is a measure against a point in time which can change in an instant. Simply put, what is read this very second might have changed in the next heart beat.

akinny (approved):

akinny (approved):

PR #65, opened 2017-04-07
Make jvm webserver delete consistent part 2

jeddcuison:
- Fix JVM not deleting when in new state in group operation page
- Fix JVM and web server dialog box not closing and refreshing the table when ok is clicked
- Wrote additional tests for new status with hard delete option
- Make sure that the JQuery selector that removes the JVM and web server from the group operations table only removes the deleted JVM or web server
- Make delete buttons only accessible by admin users

### DIFF
--- a/jwala-services/src/main/java/com/cerner/jwala/service/group/GroupStateNotificationService.java
+++ b/jwala-services/src/main/java/com/cerner/jwala/service/group/GroupStateNotificationService.java
@@ -1,6 +1,11 @@
 package com.cerner.jwala.service.group;
 
+import com.cerner.jwala.common.domain.model.group.Group;
 import com.cerner.jwala.common.domain.model.id.Identifier;
+import com.cerner.jwala.common.domain.model.state.CurrentState;
+import com.cerner.jwala.common.domain.model.state.OperationalState;
+
+import java.util.Map;
 
 /**
  * Retrieve group state details (e.g. running JVM count and Web Sever count) and send the said data to a destination via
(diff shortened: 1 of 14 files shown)


jwala-services/src/main/java/com/cerner/jwala/service/group/GroupStateNotificationService.java
```diff
@@ -1,6 +1,11 @@
 package com.cerner.jwala.service.group;
```
jeddcuison, line 1: PLEASE IGNORE THIS COMMIT. For some reason this commit (which is in another branch #63) got merged in this branch. 

Please note that since this branch only contains the impl file part of #63, doing a build with tests will fail. If you want to do a build just do gradle clean build -x test for the moment until #63 gets merge and I update this branch,

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -281,23 +281,26 @@ public void deleteJvm(final Identifier<Jvm> id, boolean hardDelete, final User u
         if (hardDelete) {
             LOGGER.info("Deleting JVM service {}", jvm.getJvmName());
 
-            if (!JvmState.JVM_STOPPED.equals(jvm.getState()) && !JvmState.FORCED_STOPPED.equals(jvm.getState())) {
+            if (!JvmState.JVM_NEW.equals(jvm.getState()) && !JvmState.JVM_STOPPED.equals(jvm.getState()) &&
+                    !JvmState.FORCED_STOPPED.equals(jvm.getState())) {
                 final String msg = MessageFormat.format("Please stop JVM {0} first before attempting to delete it",
```
Sharvaribarve, line 286: Could change the message to stop completely...considering stopping state.

jeddcuison: I see your point but for me it means the same thing. Anyways, let's just have the team vote on it.

Hi @akinny, @jmahmoodcc, @stevenger, @PeterTeunissen, @rahulsayini, @LinHungWu 

Can we get your thoughts on this ? Thanks!

akinny: I think the message is ok for now

akinny (approved):

ghost (approved):

PR #69, opened 2017-04-10
notify the UI that a web server was generated 

jmahmoodcc:
After generating a web server from the Operations tab, the state of the web server does not get updated from NEW to STOPPED. This causes problems when trying to start the web server from the UI immediately after generating.

The fix is to send the state update to the UI immediately after completing the web server generation.

### DIFF
--- a/jwala-services/src/main/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImpl.java
+++ b/jwala-services/src/main/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImpl.java
@@ -139,9 +139,10 @@ public class WebServerServiceImpl implements WebServerService {
                                      final User anUpdatingUser) {
         anUpdateWebServerCommand.validate();
 
-        WebServer originalWebServer=getWebServer(anUpdateWebServerCommand.getId());
-        if(!anUpdateWebServerCommand.getNewName().equalsIgnoreCase(originalWebServer.getName())&& !anUpdateWebServerCommand.getState().toStateLabel().equals(WebServerReachableState.WS_NEW.toStateLabel())){
-            throw new WebServerServiceException("Web Server "+originalWebServer.getName()+" is in "+originalWebServer.getState().toStateLabel()+" state, can only rename new web servers");
+        WebServer originalWebServer = getWebServer(anUpdateWebServerCommand.getId());
+        if (!anUpdateWebServerCommand.getNewName().equalsIgnoreCase(originalWebServer.getName()) && !WebServerReachableState.WS_NEW.equals(originalWebServer.getState())) {
+            throw new WebServerServiceException(MessageFormat.format("Web Server {0} is in {1} state, can only rename new web servers",
+                    originalWebServer.getName(), originalWebServer.getState().toStateLabel()));
         }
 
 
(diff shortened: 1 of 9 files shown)


akinny (approved):

jwala-services/src/main/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImpl.java
```diff
@@ -208,6 +220,7 @@ public boolean isStarted(WebServer webServer) {
     public void updateState(final Identifier<WebServer> id, final WebServerReachableState state, final String errorStatus) {
         webServerPersistenceService.updateState(id, state, errorStatus);
         inMemoryStateManagerService.put(id, state);
+        messagingTemplate.convertAndSend(topicServerStates, new CurrentState<>(id, state, DateTime.now(), StateType.WEB_SERVER));
```
jeddcuison: The WebServerStateSetterWorker setState sends a message to the UI already via the MessageService. If we put this here, the message will always be sent 2x. Can we put this after generation ? When we set the state to STOPPED ? In addition "updateState" in my opinion should just update the persistence layer and the hashmap, it's up to the client service to do notifications.

jeddcuison (changes requested):
Message is sent 2x on ping since it uses WebServerService update.

jwala-services/src/main/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImpl.java
```diff
@@ -27,10 +29,12 @@
 import com.cerner.jwala.service.webserver.exception.WebServerServiceException;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
+import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.messaging.simp.SimpMessagingTemplate;
```
jeddcuison: In case we need to send a message, can we just use the MessagingService instead ?

jwala-services/src/main/java/com/cerner/jwala/service/webserver/impl/WebServerServiceImpl.java
jmahmoodcc: Good catch! Making the changes now ...

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/webserver/impl/WebServerServiceRestImpl.java
```diff
@@ -60,6 +63,10 @@
 
     @Autowired
     BinaryDistributionLockManager binaryDistributionLockManager;
+
+    @Autowired
+    SimpMessagingServiceImpl simpleMessagingService;
```
jeddcuison: Can this be private ? Can we use the interface instead ? e.g. MessagingService

jeddcuison (approved):
Looks good, I just have a few comments - minor ones, if it's not too much hassle can you look it ? Anyways you can merge it whether you did the comments or not. :)

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/webserver/impl/WebServerServiceRestImpl.java
jmahmoodcc: Done

PR #70, opened 2017-04-11
Check the jvm state first before checking if state is stale

jeddcuison:
This prevents unnecessary logging  that the JVM is stale (stopped JVMs are supposed to be inactive so there's no point in checking if their state is stale or not)

### DIFF
--- a/jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/spring/component/JvmStateServiceImpl.java
+++ b/jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/spring/component/JvmStateServiceImpl.java
@@ -102,7 +102,7 @@ public class JvmStateServiceImpl implements JvmStateService {
         }
 
         for (final Jvm jvm : jvms) {
-            if (stateNotInMemory(jvm) || isValidState(jvm) && isStale(jvm) && isFutureNilOrDone(jvm)) {
+            if (stateNotInMemory(jvm) || isValidState(jvm) && isFutureNilOrDone(jvm) && isStale(jvm)) {
                 LOGGER.debug("Pinging JVM {} ...", jvm.getJvmName());
                 PING_FUTURE_MAP.put(jvm.getId(), jvmStateResolverWorker.pingAndUpdateJvmState(jvm, this));
                 LOGGER.debug("Pinged JVM {}", jvm.getJvmName());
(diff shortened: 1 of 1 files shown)


jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/spring/component/JvmStateServiceImpl.java
```diff
@@ -102,7 +102,7 @@ public void verifyAndUpdateJvmStates() {
         }
 
         for (final Jvm jvm : jvms) {
-            if (stateNotInMemory(jvm) || isStale(jvm) && isValidState(jvm) && isFutureNilOrDone(jvm)) {
+            if (stateNotInMemory(jvm) || isValidState(jvm) && isStale(jvm) && isFutureNilOrDone(jvm)) {
                 LOGGER.debug("Pinging JVM {} ...", jvm.getJvmName());
```
akinny: There is no change in the logic, what am I missing?

akinny (approved):

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/spring/component/JvmStateServiceImpl.java
```diff
@@ -102,7 +102,7 @@ public void verifyAndUpdateJvmStates() {
         }
 
         for (final Jvm jvm : jvms) {
-            if (stateNotInMemory(jvm) || isStale(jvm) && isValidState(jvm) && isFutureNilOrDone(jvm)) {
+            if (stateNotInMemory(jvm) || isValidState(jvm) && isStale(jvm) && isFutureNilOrDone(jvm)) {
```
jmahmoodcc: I don't understand how switching these checks make any difference - this looks like this would evaluate the same as before the change.

In Java, I thought the AND operand was always implied as being grouped together as one sub expression so it doesn't matter what order those checks are made. I ran the following unit test just to make sure:
```
    @Test
    public void testConditionals() {
        assertEquals(true, true || false && false && false);
        assertEquals(true, true || false && false && true);
        assertEquals(true, true || false && true && false);
        assertEquals(true, true || false && true && true);
        assertEquals(true, true || true && false && false);
        assertEquals(true, true || true && false && true);
        assertEquals(true, true || true && true && false);
        assertEquals(true, true || true && true && true);

        assertEquals(false, false || false && false && false);
        assertEquals(false, false || false && false && true);
        assertEquals(false, false || false && true && false);
        assertEquals(false, false || false && true && true);
        assertEquals(false, false || true && false && false);
        assertEquals(false, false || true && false && true);
        assertEquals(false, false || true && true && false);
        assertEquals(true, false || true && true && true);
    }
```
I'm not convinced that this change does anything different.

Please, persuade me otherwise!

jmahmoodcc (changes requested):
Switching those conditions for an AND operation shouldn't make a difference

jmahmoodcc (approved):

PR #71, opened 2017-04-11
Associate the "Unpack War" attribute for the Web App with the "unpack" attribute in the meta data

jmahmoodcc:
When creating a Web App, if the "Unpack War" checkbox is checked then when the war is uploaded the "unpack" attribute needs to be set in the meta data. When the Web App is updated the state of the "Unpack War" flag should be reflected in the "unpack" flag of the meta data for the war.

This should work in reverse as well: if the "unpack" flag in the meta data is updated, then the "Unpack War" flag should be in the same state for the Web App.


jwala-services/src/main/java/com/cerner/jwala/service/app/impl/ApplicationServiceImpl.java
```diff
@@ -128,7 +132,42 @@ public Application getApplication(final String name) {
...
+            try {
+                String originalJsonMetaData = groupPersistenceService.getGroupAppResourceTemplateMetaData(application.getGroup().getName(), appWarName);
+                ResourceTemplateMetaData originalMetaData = resourceService.getMetaData(originalJsonMetaData);
+                ResourceTemplateMetaData updateMetaData = new ResourceTemplateMetaData(
+                        originalMetaData.getTemplateName(),
+                        originalMetaData.getContentType(),
+                        originalMetaData.getDeployFileName(),
+                        originalMetaData.getDeployPath(),
+                        originalMetaData.getEntity(),
+                        updateApplicationRequest.isUnpackWar(),
```
akinny, line 153: What about other meta data updates, does it mean I can only update unpackwar flag?

akinny (commented):
Can you look at my comment

jwala-services/src/main/java/com/cerner/jwala/service/app/impl/ApplicationServiceImpl.java
jmahmoodcc: The unpackWar flag is the only meta data attribute that is shared by the Web App. The other Web App attributes (application name, application context path, is the application secure, and the load balance attribute) don't map to any meta data attributes. Technically, the application name and group details could be mapped to the entity attributes of the meta data, but those are no longer used for deployment or identifying the resource.

Also, this code is only called when the Web App is updated - not when the meta data is updated. So this will map the "Unpack War" attribute of the application to the "unpack" attribute of the meta data for the application war resource.

jwala-services/src/main/java/com/cerner/jwala/service/resource/impl/handler/GroupLevelAppResourceHandler.java
```diff
@@ -63,51 +72,74 @@ public ConfigTemplate fetchResource(final ResourceIdentifier resourceIdentifier)
...
+    private ResourceTemplateMetaData updateApplicationWarMetaData(ResourceIdentifier resourceIdentifier, ResourceTemplateMetaData metaDataCopy, Application app) {
+        boolean isUnpackWar = app.isUnpackWar();
+        metaDataCopy = new ResourceTemplateMetaData(metaDataCopy.getTemplateName(), metaDataCopy.getContentType(), metaDataCopy.getDeployFileName(), metaDataCopy.getDeployPath(), metaDataCopy.getEntity(), isUnpackWar, metaDataCopy.isOverwrite());
+
+        try {
+            final ObjectMapper objectMapper = new ObjectMapper();
+            String jsonData = objectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(metaDataCopy);
+            metaDataCopy.setJsonData(jsonData);
+        } catch (IOException e) {
+            final String errMsg = MessageFormat.format("Unable to convert meta data to a string for app {0} creating resource {1}", resourceIdentifier.webAppName, metaDataCopy.getDeployFileName(), e);
```
jeddcuison: For me the message that would be easier to understand in a troubleshooting scenario is the following

    Failed to update the war meta data for app {0} on {1} resource creation

The reason is because at first glance it would call my attention immediately on "update war meta data" and then for the details, I'd just read the stack trace.

Anyways this is a minor thing and might be a personal preference. The team might have another opinion.

jwala-services/src/test/java/com/cerner/jwala/service/app/impl/ApplicationServiceImplTest.java
```diff
@@ -208,13 +208,21 @@ public void testCreate() {
...
+
+        when(Config.mockGroupPersistenceService.getGroupAppResourceTemplateMetaData(anyString(), anyString())).thenReturn("{\"templateName\":\"test-template-name\", \"contentType\":\"application/zip\", \"deployFileName\":\"test-app.war\", \"deployPath\":\"/fake/deploy/path\", \"entity\":{}, \"unpack\":\"true\", \"overwrite\":\"true\"}");
+
+        when(Config.mockResourceService.getMetaData(anyString())).thenReturn(new ResourceTemplateMetaData("test-template-name", MediaType.APPLICATION_ZIP, "deploy-file-name", "deploy-path", null, true, false));
+
         UpdateApplicationRequest cac = new UpdateApplicationRequest(Config.mockApplication2.getId(), Identifier.id(1L, Group.class), "wan", "/wan", true, true, false);
         Application created = applicationService.updateApplication(cac, new User("user"));
 
         assertTrue(created == Config.mockApplication2);
+        verify(Config.mockGroupPersistenceService, times(1)).updateGroupAppResourceMetaData(anyString(), anyString(), anyString(), anyString());
```
jeddcuison: The default if you don't specify times is times(1) :)

jeddcuison (approved):
With minor comments

jwala-services/src/main/java/com/cerner/jwala/service/resource/impl/handler/GroupLevelAppResourceHandler.java
jmahmoodcc: Changed the message

jwala-services/src/test/java/com/cerner/jwala/service/app/impl/ApplicationServiceImplTest.java
jmahmoodcc: Removed the times(1)

PR #74, opened 2017-04-13
fix update jvm and webserver

rahulsayini:
The update of jvm and webserver was not allowing to update the json body. Now i fixed it by comparing the original jvm name with new jvm name.


PeterTeunissen (approved):
Leave a comment

LinHungWu:
good

LinHungWu (approved):

PR #81, opened 2017-04-18
Property and history cleanup merged with master

jeddcuison:
Add back missing STOP JVM functionality


jeddcuison:
Created another pull request since the branch this one was based on seemed problematic due additional commit "	f8c2bcf" which should not be there. Please see pull request #82

PR #83, opened 2017-04-19
Refresh group operations page after deleting a JVM or web server

jeddcuison:
This fixes the following issues:

- The deleted JVM is still in UI memory and is displayed again when user collapses and opens the row where the deleted JVM belonged to
- Group state and state count is not correct
- Web server section state count is incorrect
- JVM section state count is incorrect

Note: Refreshing the display dynamically would be very difficult without doing a lot of workarounds. The group operations page was written during the time that the team was still learning about React resulting to a complex code and ironically something that goes against some of the principals of React - my bad really. Anyways, the group operations page is really due for a refactoring. Believe it or not the rewrite is WAY less complex and probably due to experience is easier to do and finish. We just need to complete the selenium test cases for the group operations page and some time.

### DIFF
--- a/jwala-webapp/src/main/webapp/resources/js/alert.js
+++ b/jwala-webapp/src/main/webapp/resources/js/alert.js
@@ -1,10 +1,10 @@
 /**
  * sources: http://coding.abel.nu/2012/01/jquery-ui-replacement-for-alert/
  */
-$.extend({ alert: function (message, aTitle, aModal) {
+$.extend({ alert: function (message, aTitle, aModal, closeCallback) {
   $("<div></div>").dialog( {
-    buttons: { "Ok": function () { $(this).dialog("close"); } },
-    close: function (event, ui) { $(this).remove(); },
+    buttons: { "Ok": function () { $(this).dialog("close"); closeCallback(); } },
+    close: function (event, ui) { $(this).remove();  closeCallback(); },
     resizable: true,
     title: aTitle === undefined ? "Message" : aTitle,
     modal: aModal === undefined ? true : aModal
--- a/jwala-webapp/src/main/webapp/resources/js/react/group-operations/GroupOperationsDataTable.js
+++ b/jwala-webapp/src/main/webapp/resources/js/react/group-operations/GroupOperationsDataTable.js
@@ -815,12 +815,8 @@ var GroupOperationsDataTable = React.createClass({
     webServerConfirmDeleteCallback: function(data) {
         let self = this;
         webServerService.deleteWebServer(data.id.id, true).then(function(e){
-            // Remove deleted row
-             $("[id*='ws-child-table']").find("td:contains('" + data.name + "')").filter(function() {
-                if ($(this).text() === data.name) {
-                    $(this).parent().remove();
-                }
-            });
+            $.alert("Web server " + data.name + " was successfully deleted. Jwala will need to refresh to display the latest data and recompute the states.",
+                    "Delete Web Server", true, function(){location.reload()});
         }).caught(
             function(e) {
                 self.setState({showDeleteConfirmDialog: false})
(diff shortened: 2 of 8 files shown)


jwala-webapp/src/main/webapp/resources/js/react/group-operations/GroupOperationsDataTable.js
```diff
@@ -815,12 +815,8 @@ var GroupOperationsDataTable = React.createClass({
     webServerConfirmDeleteCallback: function(data) {
         let self = this;
         webServerService.deleteWebServer(data.id.id, true).then(function(e){
-            // Remove deleted row
-             $("[id*='ws-child-table']").find("td:contains('" + data.name + "')").filter(function() {
-                if ($(this).text() === data.name) {
-                    $(this).parent().remove();
-                }
-            });
+            $.alert("Web server " + data.name + " was successfully deleted. Jwala will need to refresh to display the latest data and recompute the states.",
```
jmahmoodcc, line 818: I know the trouble you went through to make this work right, but is there any other to do this without reloading the page and showing a warning that the page will be refreshed? Just seems like a messy way to do it ...

jwala-webapp/src/main/webapp/resources/js/react/group-operations/GroupOperationsDataTable.js
```diff
@@ -843,12 +839,8 @@ var GroupOperationsDataTable = React.createClass({
     jvmConfirmDeleteCallback: function(data) {
         let self = this;
         jvmService.deleteJvm(data.id.id, true).then(function(e){
-            // Remove deleted row
-            $("[id*='jvm-child-table']").find("td:contains('" + data.jvmName + "')").filter(function() {
-                if ($(this).text() === data.jvmName) {
-                    $(this).parent().remove();
-                }
-            });
+            $.alert("JVM " + data.jvmName + " was successfully deleted. Jwala will need to refresh to display the latest data and recompute the states.",
```
akinny, line 842: I don't think we need this message, most UI's refresh after such operations
"Jwala will need to refresh to display the latest data and recompute the states."

akinny (commented):
update the message, it will confuse the users

jwala-webapp/src/main/webapp/resources/js/react/group-operations/GroupOperationsDataTable.js
jmahmoodcc, line 842: Again, I wouldn't show the message. Or maybe if you have to reload, can you just refresh the container where the table exists and show the spinning gears?

jwala-webapp/src/main/webapp/resources/js/alert.js
```diff
@@ -1,10 +1,10 @@
 /**
  * sources: http://coding.abel.nu/2012/01/jquery-ui-replacement-for-alert/
  */
-$.extend({ alert: function (message, aTitle, aModal) {
+$.extend({ alert: function (message, aTitle, aModal, closeCallback) {
   $("<div></div>").dialog( {
-    buttons: { "Ok": function () { $(this).dialog("close"); } },
-    close: function (event, ui) { $(this).remove(); },
+    buttons: { "Ok": function () { $(this).dialog("close"); closeCallback(); } },
```
jmahmoodcc, line 6: is the clockCallback a required parameter now? Do we need to check if it's undefined before calling it?

jmahmoodcc (changes requested):
Can we try and see if it's a lot of work to not reload the whole page?

jwala-webapp/src/main/webapp/resources/js/alert.js
jeddcuison: It's not required. Btw it's closeCallback. ;)

jwala-webapp/src/main/webapp/resources/js/react/group-operations/GroupOperationsDataTable.js
jeddcuison: There is, but I need to refactor it. Doing it with workarounds would take the same time in my calculations than just rewriting it taking advantage of React. React is perfect for this scenario but sad to say React is not mostly at play here.

jwala-webapp/src/main/webapp/resources/js/react/group-operations/GroupOperationsDataTable.js
jeddcuison: I can't just refresh the JVMs since the JVM is attached to the group data unlike Web Server. I wanted to make JVMs load like Web Server before but the team decided not to then. I can convert it to the way Web Server does it but it's going to take a lot of work - so might as well refactor and write it the right way this time. I tried doing it just to gauge the time it will take, surprisingly I was able to refactor the group table in 2 hours or less and the code looks great. I didn't know why that one escaped me before - probably due to the learning curve.

jmahmoodcc (approved):
After discussing with the team and Jedd, I am ok with making the real changes later.

PR #86, opened 2017-04-20
Fix search webapp ui

rahulsayini:
add a if check to rdatatable.js and flag isVisible: false to webAppConfig.js


jeddcuison (changes requested):
I wanted the change to be confined in RDataTable so that if in the future some other part of Jwala uses it and passes out-of-spec data, RDataTable can handle it and you won't have to fix the other parts of Jwala. Please change RDataTable with this code:

    if (tmpVal && (col.isVisible === undefined || col.isVisible)) {
        var tmpStr = tmpVal.toString();
        ...           
    }

In addition, please write a Selenium test case for this. The test case should do the following:

1. Add at least 2 web applications
2. Do a search so much so that the 1st web app is the only one shown
3. Do a search so much so that the 2nd web app is the only one shown
4. Do a search so that 2 of the web apps are shown
5. Do a search in which the result would be no web app displayed
6. Delete the web app you added - cleanup

Note: When creating the web application, follow the way a web app is created in this test - AppCreateTest. JwalaChromeTestSuite is the test that you should run. Change test.properties user and password to n9sf user-password. Do not check in test.properties.

jwala-webapp/src/main/webapp/resources/js/react/WebAppConfig.js
```diff
@@ -20,7 +20,7 @@ var WebAppConfig = React.createClass({
                    <RDataTable ref="dataTable"
                                tableIndex="id.id"
                                colDefinitions={[{key: "id", isVisible: false},
-                                                {key: "secure", renderCallback: this.renderSecureColumn, sortable: false},
+                                                {key: "secure", renderCallback: this.renderSecureColumn, sortable: false, isVisible: false},
```
jeddcuison: Can you revert this ? Because we don't need this change anymore. Remember one of the reasons for the change request is to limit the code change in RDataTable.js ? To revert it checkout from master again to a new branch e.g. fix-search-webapp-ui-2, copy the change you did for RDataTable.js, push then create another pull request. Afterwards close this pull request putting in the comment that you have a new pull request. Put the number of the new pull request in the said comment.

jeddcuison (changes requested):
Please revert the change to WebAppConfig.js since we don't need to change that.

akinny (approved):

jmahmoodcc (changes requested):
Jedd requested a selenium test be written as well. Please create the test that Jedd described.

PeterTeunissen (approved):

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/testsuite/configuration/app/AppSearchTest.java
```diff
@@ -0,0 +1,64 @@
...
+        driver.switchTo().activeElement().sendKeys(APP_NAME2);
+        driver.switchTo().activeElement().sendKeys(Keys.TAB);
+        driver.switchTo().activeElement().sendKeys("/zApp2");
+        driver.findElement(By.xpath("//div[contains(text(), '" + GROUP_NAME_1 + "')]/input")).click();
+        driver.findElement(By.xpath("//button[span[text()='Ok']]")).click();
+
+        Thread.sleep(1000);
+        driver.findElement(By.xpath("//span[contains(text(), 'Search')]/following-sibling::input")).click();
+        driver.switchTo().activeElement().sendKeys("zApp1");
+        Thread.sleep(1000);
```
jmahmoodcc: Where do you validate that the search results contain the expected app?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/testsuite/configuration/app/AppSearchTest.java
```diff
@@ -0,0 +1,64 @@
...
+        driver.findElement(By.xpath("//span[contains(text(), 'Search')]/following-sibling::input")).clear();
+        driver.switchTo().activeElement().sendKeys("zApp2");
```
jmahmoodcc, line 72: Same thing: where is the validation of the search results?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/testsuite/configuration/app/AppSearchTest.java
```diff
@@ -0,0 +1,64 @@
...
+        driver.findElement(By.xpath("//span[contains(text(), 'Search')]/following-sibling::input")).clear();
+        Thread.sleep(1000);
+        driver.switchTo().activeElement().sendKeys("zApp");
```
jmahmoodcc: Again, where is the validation of the search results?

jmahmoodcc (changes requested):
Maybe I missed, but I didn't see where the search result is validated in the selenium test

LinHungWu (approved):

rahulsayini:
opened a new branch UI-webapp-search https://github.com/cerner/jwala/pull/111

PR #91, opened 2017-05-02
Added validation for length of JVM name

Sharvaribarve:
Checking if JVM names are less than 65 characters/

### DIFF
--- a/jwala-common/src/main/java/com/cerner/jwala/common/properties/PropertyKeys.java
+++ b/jwala-common/src/main/java/com/cerner/jwala/common/properties/PropertyKeys.java
@@ -33,7 +33,7 @@ public enum PropertyKeys {
     PRIVATE_KEY_FILE("ssh.privateKeyFile"),
     KNOWN_HOSTS_FILE("ssh.knownHostsFile"),
     ENCRYPTED_PASSWORD("ssh.encrypted.password"),
-    MOD_PROXY_OVERRIDE("mod_proxy_override");
+    JVM_ROUTE_MAX_LENGTH("jvm.route.max.length");
 
     private String propertyName;
 
--- a/jwala-common/src/main/java/com/cerner/jwala/common/properties/PropertyKeys.java
+++ b/jwala-common/src/main/java/com/cerner/jwala/common/properties/PropertyKeys.java
@@ -32,7 +32,8 @@ public enum PropertyKeys {
     PORT("ssh.port"),
     PRIVATE_KEY_FILE("ssh.privateKeyFile"),
     KNOWN_HOSTS_FILE("ssh.knownHostsFile"),
-    ENCRYPTED_PASSWORD("ssh.encrypted.password");
+    ENCRYPTED_PASSWORD("ssh.encrypted.password"),
+    MOD_PROXY_OVERRIDE("mod_proxy_override");
 
     private String propertyName;
 
--- a/jwala-common/src/main/java/com/cerner/jwala/common/rule/jvm/JvmNameRule.java
+++ b/jwala-common/src/main/java/com/cerner/jwala/common/rule/jvm/JvmNameRule.java
@@ -16,8 +16,13 @@ public class JvmNameRule extends ValidNameRule {
     }
 
     @Override
+    public boolean isValid() {
+        return super.isValid() && name.length() < 65;
+    }
+
+    @Override
     protected String getMessage() {
-        return "Invalid Jvm Name : \"" + name + "\"";
+        return name != null && name.length() > 64 ? "Length of JVM should not exceed 64 characters" : "Invalid Jvm Name : \"" + name + "\"";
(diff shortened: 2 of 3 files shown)


jwala-common/src/main/java/com/cerner/jwala/common/rule/jvm/JvmNameRule.java
```diff
@@ -24,7 +24,7 @@ public boolean isValid() {
 
     @Override
     protected String getMessage() {
-        return name != null && name.length() > MAX_LENGTH ? "Length of JVM should not exceed 64 characters" : "Invalid Jvm Name : \"" + name + "\"";
+        return name != null && name.length() > MAX_LENGTH ? "Length of JVM should not exceed "+ MAX_LENGTH+ " characters" : "Invalid Jvm Name : \"" + name + "\"";
```
jeddcuison: Please put spaces in between operators e.g.

    "blah blah..." + CONSTANT + "blah blah..."

jeddcuison: Please wrap the line if it's too long. 125 characters is the de facto standard right now so what I do is set my editor's guideline at 120 character so I would know when to wrap. If it exceeds a little from 120 it is fine.

image

jeddcuison (commented):
I have some minor comments

jmahmoodcc (changes requested):
Maybe I missed it: where is the rule applied?
Also: do we need to do the same thing for Web Server names?

jwala-common/src/main/java/com/cerner/jwala/common/properties/PropertyKeys.java
```diff
@@ -32,7 +32,8 @@
     PORT("ssh.port"),
     PRIVATE_KEY_FILE("ssh.privateKeyFile"),
     KNOWN_HOSTS_FILE("ssh.knownHostsFile"),
-    ENCRYPTED_PASSWORD("ssh.encrypted.password");
+    ENCRYPTED_PASSWORD("ssh.encrypted.password"),
+    MOD_PROXY_OVERRIDE("mod_proxy_override");
```
PeterTeunissen: this is a bad name for the property... should be renamed to something like "jvm.route.max.length"

PeterTeunissen (changes requested):
Seem comment: property name does not reflect what it does.

jwala-common/src/main/java/com/cerner/jwala/common/properties/PropertyKeys.java
jmahmoodcc: +1

Also make sure to follow the syntax conventions of the other properties: all lower case and separated by periods

jwala-common/src/main/java/com/cerner/jwala/common/rule/jvm/JvmNameRule.java
```diff
@@ -1,11 +1,16 @@
...
+import com.cerner.jwala.common.domain.model.app.Application;
 import com.cerner.jwala.common.domain.model.fault.FaultType;
 import com.cerner.jwala.common.exception.MessageResponseStatus;
+import com.cerner.jwala.common.properties.ApplicationProperties;
+import com.cerner.jwala.common.properties.PropertyKeys;
 import com.cerner.jwala.common.rule.ValidNameRule;
 
 public class JvmNameRule extends ValidNameRule {
 
+    public final int MAX_LENGTH = ApplicationProperties.getAsInteger(PropertyKeys.MOD_PROXY_OVERRIDE, 64);
```
jmahmoodcc: Rename the key to something like Peter suggested: JVM_ROUTE_MAX_LENGTH

jwala-common/src/main/java/com/cerner/jwala/common/rule/jvm/JvmNameRule.java
```diff
@@ -16,8 +21,13 @@ protected MessageResponseStatus getMessageResponseStatus() {
...
 
     @Override
+    public boolean isValid() {
+        return super.isValid() && name.length() <= MAX_LENGTH;
+    }
+
+    @Override
     protected String getMessage() {
-        return "Invalid Jvm Name : \"" + name + "\"";
+        return name != null && name.length() > MAX_LENGTH ? "Length of JVM should not exceed " + MAX_LENGTH + " characters" : "Invalid Jvm Name : \"" + name + "\"";
```
jmahmoodcc: I would add instructions on how to increase the max length using the new property here as well - just in case the client is using a load balancer that supports more than 64 characters for the route name

jmahmoodcc (changes requested):
See comments about naming of the property and make the property known to the user when the JVM name exceeds the max length

akinny (approved):

PR #79, opened 2017-04-17
History logging of resources at web app

Sharvaribarve:
Added history statement for deploying of resources at webapp level
Before you contribute, please review these guidelines to help ensure a smooth process for everyone.

Thanks.

* Read [how to properly contribute to open source projects on Github][1].
* Fork the project.
* Use a feature branch.
* Write [good commit messages][2].
* Use the same coding conventions as the rest of the project.
* Commit locally and push to your fork until you are happy with your contribution.
* Make sure to add tests and verify all the tests are passing when merging upstream.
* Add an entry to the [Changelog][3] accordingly.
* Please add your name to the [CONTRIBUTORS.md][7] file. Adding your name to the [CONTRIBUTORS.md][7] file signifies agreement to all rights and reservations provided by the [License][4].
* [Squash related commits together][5].
* Open a [pull request][6].
* The pull request will be reviewed by the community and merged by the project committers.

[1]: http://gun.io/blog/how-to-github-fork-branch-and-pull-request
[2]: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[3]: ./CHANGELOG.md
[4]: ./LICENSE
[5]: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html
[6]: https://help.github.com/articles/using-pull-requests
[7]: ./CONTRIBUTORS.md

PR #92, opened 2017-05-04
Resources history how it should be

jeddcuison:
This pull request is based on ResourcesHistory (with pull request #79 ). I noticed that ResourcesHistory has a commit with the title "Merge branch 'master' of https://github.com/cerner/jwala into ResourcesHistory". The way to update the branch without adding another commit is by doing this

    git pull --rebase origin master // one should be in their branch, if there are conflicts it should be resolved in Intellij unless one uses Eclipse then I guess that's a different story

Anyways, the problem that I had with the "Merge branch 'master'..." commit is that the changes that really are of the branch is not reflected in the said commit. One would only see the changes made in master which I think should not be the case.

If reviewers would check the commits in this branch, they would only see changes meant for this branch which makes it easier for them to read. In addition if the pull request is merged to master (squashed or not) then the commit history would look clean.


akinny (approved):

jmahmoodcc (changes requested):
See comments in https://github.com/cerner/jwala/pull/79

jeddcuison:
This pull request is for example purposes

PR #93, opened 2017-05-04
check for existing group name for create and update operations

rahulsayini:


jwala-services/src/main/java/com/cerner/jwala/service/group/impl/GroupServiceImpl.java
```diff
@@ -65,7 +67,14 @@ public GroupServiceImpl(final GroupPersistenceService groupPersistenceService,
...
-
+        try {
+            groupPersistenceService.getGroup(createGroupRequest.getGroupName());
+            String message = MessageFormat.format("Group Name already exists: {0} ", createGroupRequest.getGroupName());
+            LOGGER.error(message);
+            throw new EntityExistsException(message);
+        } catch (NotFoundException e) {
+            LOGGER.debug("No group name conflict, ignoring not found exception for creating group ", e);
+        }
         createGroupRequest.validate();
```
jeddcuison: Does validate check if the group name is empty or has illegal characters ? If so, createGroupRequest.validate() should go first before you check for duplicates.

jwala-services/src/main/java/com/cerner/jwala/service/group/impl/GroupServiceImpl.java
```diff
@@ -112,6 +121,16 @@ public Group getGroup(final String name) {
...
+        try {
+            if (!orginalGroup.getName().equalsIgnoreCase(anUpdateGroupRequest.getNewName()) && null != groupPersistenceService.getGroup(anUpdateGroupRequest.getNewName())) {
+                String message = MessageFormat.format("Group Name already exists: {0}", anUpdateGroupRequest.getNewName());
+                LOGGER.error(message);
+                throw new EntityExistsException(message);
+            }
+        } catch (NotFoundException e) {
+            LOGGER.debug("No group name conflict, ignoring not found exception for creating group ", e);
+        }
         anUpdateGroupRequest.validate();
```
jeddcuison: If validate checks if the name is legal e.g. is null, has illegal characters, length is good etc then validate should go first before checking for duplicates

jwala-services/src/test/java/com/cerner/jwala/service/group/impl/GroupServiceImplVerifyTest.java
```diff
@@ -59,17 +61,28 @@ public GroupServiceImplVerifyTest() {
...
     @Test
     public void testCreateGroup() {
 
-        final CreateGroupRequest command = mock(CreateGroupRequest.class);
+        final CreateGroupRequest command = new CreateGroupRequest("Group");
+        when(Config.mockGroupPersistenceService.getGroup(anyString())).thenThrow(NotFoundException.class);
+        groupService.createGroup(command,
+                user);
+
+        verify(Config.mockGroupPersistenceService, times(1)).createGroup(command);
```
jeddcuison, line 75: This is minor, just an FYI - verify's default Times is 1 so you can omit it in the future

    verify(someMock).doSomething(...)

jeddcuison (commented):
If validate checks if group name is empty, has illegal characters or has invalid length then it should execute first before checking for duplicates

PeterTeunissen (approved):

jmahmoodcc (approved):

PR #96, opened 2017-05-09
Error Message on clicking show details should display detailed error message instead of [object object]

Sharvaribarve:
In case, the validation of resource file fails, the show details button on clicking displays [object object], this has been changed to display detailed error message.

### DIFF
--- a/jwala-webapp/src/main/webapp/resources/js/error-alert.js
+++ b/jwala-webapp/src/main/webapp/resources/js/error-alert.js
@@ -46,6 +46,14 @@ $.extend({ errorAlert: function (message, dlgTitle, modal, content) {
 
             var detailsHtml = "";
             if (content && content !== "") {
+            let details = content;
+                 if(!Array.isArray(content)){
+                 details = new Array(content.length);
+                 for(let key in content){
+                 details.push(content[key]);
+                 }
+                 }
+                content = details;
                 var showErrorDetails = "$('.showErrorDetails').hide();$('.hideErrorDetails').show();$('.stackTrace').show()";
                 var hideErrorDetails = "$('.showErrorDetails').show();$('.hideErrorDetails').hide();$('.stackTrace').hide()";
                 detailsHtml = "<br><br><div style='width:100%;text-align:left'><button class='showErrorDetails' onClick="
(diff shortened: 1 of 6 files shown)


jmahmoodcc (changes requested):
Looks good, just clean up the formatting please (the indenting is off)

jeddcuison (changes requested):
In lieu of Sharvari's findings that the details are the same as the message content displayed by the error dialog, the solution as discussed with the team is just to display the message. My bad I did not see this coming, the details section is supposed to be for stacktrace details. Please revert the changes and just omit the content parameter in $.errorAlert. By not providing the content, errorAlert is smart enough not to display the details button. Thanks!

jwala-webapp/src/main/webapp/resources/js/react/ResourcePane.js
```diff
@@ -166,7 +166,7 @@ var ResourcePane = React.createClass({
         }
     },
     displayDeployErrorMessage: function(response){
-        $.errorAlert(ResourcePane.parseDetailedErrorMsg(response, ResourcePane.DEFAULT_DEPLOY_ERR_MSG), "", undefined, response.responseJSON.content);
+        $.errorAlert(ResourcePane.parseDetailedErrorMsg(response, ResourcePane.DEFAULT_DEPLOY_ERR_MSG), "");
```
jeddcuison: Please revert to:

    $.errorAlert(ResourcePane.parseDetailedErrorMsg(response, ResourcePane.DEFAULT_DEPLOY_ERR_MSG), "", undefined, response.responseJSON.content);

jwala-webapp/src/main/webapp/resources/js/error-alert.js
```diff
@@ -45,15 +45,7 @@ $.extend({ errorAlert: function (message, dlgTitle, modal, content) {
...
-            if (content && content !== "") {
-                let details = content;
-                if (!Array.isArray(content)) {
-                    details = new Array(content.length);
-                    for (let key in content) {
-                        details.push(content[key] +"<br/>");
-                    }
-                }
-                content = details;
+            if (content && content !== "" && Array.isArray(content)) {
```
jeddcuison: Please remove unnecessary content validation, this will do 

    if (Array.isArray(content)) {
        ...
    }

jeddcuison (changes requested):
Almost there, 2 additional comments, after those are addressed then we're good to go

jeddcuison (commented):
The last change looks good. Can you write a selenium test for this ?

jmahmoodcc (approved):

LinHungWu (approved):

PR #100, opened 2017-05-10
Check for duplicate names in media

rahulsayini:


jwala-persistence/src/main/java/com/cerner/jwala/persistence/jpa/domain/JpaMedia.java
```diff
@@ -21,7 +21,7 @@
  * Created by Jedd Cuison on 12/6/2016
  */
 @Entity(name = "media")
-@NamedQueries({@NamedQuery(name = JpaMedia.QUERY_FIND_BY_NAME, query = "SELECT m FROM media m WHERE m.name = :name")})
+@NamedQueries({@NamedQuery(name = JpaMedia.QUERY_FIND_BY_NAME, query = "SELECT m FROM media m WHERE lower(m.name) = lower(:name)")})
```
PeterTeunissen: Should this query not also include the media type? Looking at this query you can not have a JDK with name 'Peter' and a Tomcat with 'Peter' ??

PeterTeunissen (changes requested):

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -97,12 +105,13 @@ public void remove(final String name) {
...
     /**
      * This method will check for the existing jvm associations for the media
+     *
      * @param name media name
      */
     private void checkForJvmAssociation(String name) {
         List<Jvm> jvmList = jvmPersistenceService.getJvms();
         for (Jvm jvm : jvmList) {
-            if (name.equalsIgnoreCase(jvm.getJdkMedia().getName())) {
+            if (jvm.getJdkMedia() != null && name.equalsIgnoreCase(jvm.getJdkMedia().getName())) {
```
akinny: Can you add the JVM name(s) in the message? Add the JVM Names in a List and throw an exception outside the loop. This will help the users identify the jvm associated with the media and delete it or change them if required.

akinny (commented):
Can you make the change I noticed?

jwala-persistence/src/main/java/com/cerner/jwala/persistence/jpa/domain/JpaMedia.java
jmahmoodcc: Agreed that we uniqueness should be based on type

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
jmahmoodcc: +1

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -97,12 +105,13 @@ public void remove(final String name) {
 
     /**
      * This method will check for the existing jvm associations for the media
+     *
      * @param name media name
      */
     private void checkForJvmAssociation(String name) {
```
jmahmoodcc: Where is this method used? Should it be removed?

jmahmoodcc (changes requested):
See comments about changes requested

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
rahulsayini: Yes, we are using this method to check the jvm association (please check line 106)

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -78,10 +83,12 @@ public JpaMedia create(final Map<String, String> mediaDataMap, final Map<String,
         final String filename = Paths.get((String) mediaFileDataMap.get("filename")).getFileName().toString();
 
         try {
-            mediaDao.find(media.getName());
-            final String msg = MessageFormat.format("Media already exists with name {0}", media.getName());
-            LOGGER.error(msg);
-            throw new MediaServiceException(msg);
+            JpaMedia media1 = mediaDao.find(media.getName());
```
jmahmoodcc: Do not name variables by adding numbers on the end. Please rename media1 to something more meaningful,  such as existingMedia.

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -78,10 +83,12 @@ public JpaMedia create(final Map<String, String> mediaDataMap, final Map<String,
...
+            if (media.getType().equals(media1.getType())) {
```
jmahmoodcc: Why not use the new findByNameAndType method here?

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -78,10 +83,12 @@ public JpaMedia create(final Map<String, String> mediaDataMap, final Map<String,
...
+                final String msg = MessageFormat.format("Media already exists with name {0}", media.getName());
```
jmahmoodcc: Add the media type in the message as well

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -113,16 +120,28 @@ public void remove(final String name) {
...
      */
-    private void checkForJvmAssociation(String name) {
+    private void checkForAssociation(String name, MediaType type) {
         List<Jvm> jvmList = jvmPersistenceService.getJvms();
+        List<WebServer> webServerList = webServerPersistenceService.getWebServers();
         List<String> existingAssosiations = new ArrayList<>();
         for (Jvm jvm : jvmList) {
-            if (jvm.getJdkMedia() != null && name.equalsIgnoreCase(jvm.getJdkMedia().getName())) {
-                existingAssosiations.add(jvm.getJvmName());
+            if (jvm.getJdkMedia().getType() == type || jvm.getTomcatMedia().getType() == type) {
```
jmahmoodcc: Personally, I would group all of the jdkMedia and jdkTomcat conditional tests together. For instance, if the type is Tomcat, the code will still check for jdkMedia and jdkTomcat - and now that I think about it, won't this add existing associations more than once? I would change existingAssociations to a Set as well to avoid duplicates.

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -138,8 +157,7 @@ private void checkForJvmAssociation(String name) {
     public JpaMedia update(final JpaMedia media) {
         JpaMedia originalMedia = mediaDao.findById(media.getId());
         try {
-            mediaDao.find(media.getName());
-            if (!originalMedia.getName().equalsIgnoreCase(media.getName())) {
+            if (!originalMedia.getName().equalsIgnoreCase(media.getName()) && !originalMedia.getType().equals(media.getType())) {
```
jmahmoodcc: What if the name isn't being updated? Won't update fail?

jwala-services/src/test/java/com/cerner/jwala/service/impl/MediaServiceImplTest.java
```diff
@@ -197,6 +204,7 @@ public void testGetMediaTypes() {
         private static final RepositoryService MOCK_MEDIA_REPOSITORY_SERVICE = mock(RepositoryService.class);
```
jmahmoodcc: I don't think these should be capitalized since they're mock instantiations of service classes, and not really enums or constants

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/media/impl/MediaServiceRestImpl.java
```diff
@@ -71,9 +72,22 @@ public Response updateMedia(final JpaMedia media, final AuthenticatedUser aUser)
     }
 
     @Override
-    public Response removeMedia(final String name, final AuthenticatedUser aUser) {
+    public Response removeMedia(final String name, String type, final AuthenticatedUser aUser) {
+        MediaType mediaType = null;
+        if (type == null) {
```
jmahmoodcc: SInce type is a required PathParam I would throw an exception here. In general we don't want to set defaults for parameters that aren't provided or don't match our enums.

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/media/impl/MediaServiceRestImpl.java
```diff
@@ -71,9 +72,22 @@ public Response updateMedia(final JpaMedia media, final AuthenticatedUser aUser)
...
+            mediaType = MediaType.JDK;
+        }
+        if (type.contains("JDK")) {
```
jmahmoodcc: Don't test with a String here, we have the enum already defined with the string value so use that instead.
`MediaType.JDK.getDisplayName().equals(type)`

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/media/impl/MediaServiceRestImpl.java
```diff
@@ -71,9 +72,22 @@ public Response updateMedia(final JpaMedia media, final AuthenticatedUser aUser)
...
+            mediaType = MediaType.JDK;
+        }
+        if (type.contains("HTTPD")) {
```
jmahmoodcc: Same thing, don't test with the String "HTTPD"

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/media/impl/MediaServiceRestImpl.java
```diff
@@ -71,9 +72,22 @@ public Response updateMedia(final JpaMedia media, final AuthenticatedUser aUser)
...
+            mediaType = MediaType.APACHE;
+        }
+        if (type.contains("Tomcat")) {
```
jmahmoodcc: Same thing: don't use the "Tomcat" String

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/media/impl/MediaServiceRestImpl.java
```diff
@@ -94,8 +108,7 @@ public Response getMediaTypes() {
         return ResponseBuilder.ok(mediaService.getMediaTypes());
     }
 
-    public void setMediaService(MediaService mediaService)
-    {
+    public void setMediaService(MediaService mediaService) {
```
jmahmoodcc: Why do we have this method?
If the mediaService isn't already autowired please use autowire and remove this setter.

jmahmoodcc (changes requested):
Some more minor changes

jmahmoodcc (approved):

PR #101, opened 2017-05-11
update JvmStateReceiverAdapter to be backwards compatible with older ReportingLifeCycleListener

jmahmoodcc:
In the older versions of the ReportingLifeCycleListener the JVM state message came in the format of key/value pairs for the following info:
 - ID => ReportingJmsMessageKey/String
 - STATE => ReportingJmsMessageKey/String

And for the latest ReportingLifeCycleListener the JVM state message came in the format of:
 - ID => String/String
 - NAME => String/String
 - STATE => String/TomcatLifecycleState 

So, in older versions of jwala that are upgraded to the latest, the application JVMs will still send their state messages in the old format which is not handled by the latest JvmStateReceiverAdapter

### DIFF
--- a/jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
+++ b/jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
@@ -79,14 +79,14 @@ public class JvmStateReceiverAdapter extends ReceiverAdapter {
             return null;
         }
 
-        final Object initialKey = keys.iterator().next();
-        if (initialKey instanceof String) {
+        if (serverInfoMap.containsKey(STATE_KEY)) {
             // convert the LifecycleState to the JvmState
             return LIFECYCLE_JWALA_JVM_STATE_REF_MAP.get(serverInfoMap.get(STATE_KEY));
         }
 
         // assume the key is ReportingJmsMessageKey, in which case the value is already returned as a string JvmState
         try {
+            final Object initialKey = keys.iterator().next();
             final Field idKey = initialKey.getClass().getDeclaredField(STATE_KEY);
             final String jvmStateString = (String) serverInfoMap.get(idKey.get(initialKey));
             return JvmState.valueOf(jvmStateString);
@@ -122,13 +122,13 @@ public class JvmStateReceiverAdapter extends ReceiverAdapter {
             return null;
         }
 
-        final Object initialKey = keys.iterator().next();
-        if (initialKey instanceof String) {
+        if (serverInfoMap.containsKey(key)) {
             return (String) serverInfoMap.get(key);
         }
 
         // assume the key is ReportingJmsMessageKey
         try {
+            final Object initialKey = keys.iterator().next();
             final Field idKey = initialKey.getClass().getDeclaredField(key);
             return (String) serverInfoMap.get(idKey.get(initialKey));
         } catch (NoSuchFieldException e) {
--- a/jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
+++ b/jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
@@ -15,10 +15,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.persistence.NoResultException;
+import java.lang.reflect.Field;
 import java.util.HashMap;
 import java.util.Map;
-
-import static org.apache.cxf.interceptor.LoggingMessage.ID_KEY;
+import java.util.Set;
 
 /**
  * The listener for JGroup messages
@@ -26,8 +26,9 @@ import static org.apache.cxf.interceptor.LoggingMessage.ID_KEY;
 public class JvmStateReceiverAdapter extends ReceiverAdapter {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(JvmStateReceiverAdapter.class);
-    public static final String STATE_KEY = "STATE";
-    public static final String NAME_KEY = "NAME";
+    static final String STATE_KEY = "STATE";
+    static final String NAME_KEY = "NAME";
+    private static final String ID_KEY = "ID";
 
     private final JvmStateService jvmStateService;
     private final JvmPersistenceService jvmPersistenceService;
@@ -60,7 +61,7 @@ public class JvmStateReceiverAdapter extends ReceiverAdapter {
     public void receive(final Message jGroupMsg) {
         final Map serverInfoMap = (Map) jGroupMsg.getObject();
         final Jvm jvm = getJvm(serverInfoMap);
-        final JvmState jvmState = LIFECYCLE_JWALA_JVM_STATE_REF_MAP.get(serverInfoMap.get(STATE_KEY));
+        final JvmState jvmState = getJvmState(serverInfoMap);
 
         if (jvm != null && !JvmState.JVM_STOPPED.equals(jvmState)) {
             jvmStateService.updateState(jvm, jvmState, StringUtils.EMPTY);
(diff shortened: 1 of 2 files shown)


jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
```diff
@@ -69,18 +70,73 @@ public void receive(final Message jGroupMsg) {
         }
     }
 
+    private JvmState getJvmState(final Map serverInfoMap) {
+
+        // first check if the key types are string
+        final Set keys = serverInfoMap.keySet();
+        if (keys.isEmpty()) {
```
akinny: Can you add a null check keys != null  ? Might cause NPE

jeddcuison: Better yet you can do this

    if (CollectionUtils.isEmpty(keys)) {
            return null;
    }

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
```diff
@@ -69,18 +70,73 @@ public void receive(final Message jGroupMsg) {
...
         // try to get the JVM by name instead
-        return serverInfoMap.get(NAME_KEY) instanceof String ? getJvmByName((String) serverInfoMap.get(NAME_KEY)) : null;
+        return getJvmByName(getStringFromMessageMap(serverInfoMap, NAME_KEY));
+    }
+
+    private String getStringFromMessageMap(final Map serverInfoMap, final String key) {
+
+        // first check if the key types are string
+        final Set keys = serverInfoMap.keySet();
+        if (keys.isEmpty()) {
```
jeddcuison: To avoid NPE you can use CollectionUtils.isEmpty

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
```diff
@@ -69,18 +70,73 @@ public void receive(final Message jGroupMsg) {
...
+        if (initialKey instanceof String) {
+            return (String) serverInfoMap.get(key);
+        }
+
+        // assume the key is ReportingJmsMessageKey
+        try {
+            final Field idKey = initialKey.getClass().getDeclaredField(key);
+            return (String) serverInfoMap.get(idKey.get(initialKey));
+        } catch (NoSuchFieldException e) {
+            LOGGER.error("Failed to find key " + key +" as ReportingJmsMessageKey in message: {}", serverInfoMap, e);
```
jeddcuison: Can be written as

    LOGGER.error("Failed to find key {} as ReportingJmsMessageKey in message: {}", key, serverInfoMap, e);

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
```diff
@@ -69,18 +70,73 @@ public void receive(final Message jGroupMsg) {
...
+            return null;
+        } catch (IllegalAccessException e) {
+            LOGGER.error("Failed to convert " + key + " key to ReportingJmsMessageKey in message: {}", serverInfoMap, e);
```
jeddcuison: Can be written as

    LOGGER.error("Failed to convert {} key to ReportingJmsMessageKey in message: {}", key, serverInfoMap, e);

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
```diff
@@ -69,18 +70,73 @@ public void receive(final Message jGroupMsg) {
...
+        final String id = getStringFromMessageMap(serverInfoMap, ID_KEY);
         if (id != null && NumberUtils.isNumber(id)) {
             return getJvmById(Long.parseLong(id));
         }
         // try to get the JVM by name instead
-        return serverInfoMap.get(NAME_KEY) instanceof String ? getJvmByName((String) serverInfoMap.get(NAME_KEY)) : null;
+        return getJvmByName(getStringFromMessageMap(serverInfoMap, NAME_KEY));
+    }
+
+    private String getStringFromMessageMap(final Map serverInfoMap, final String key) {
```
jeddcuison, line 118: Can we rename this to getJvmIdString ?

jeddcuison (commented):
Can we write a unit test for this change ?

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
```diff
@@ -69,18 +70,73 @@ public void receive(final Message jGroupMsg) {
...
+    private JvmState getJvmState(final Map serverInfoMap) {
+
+        // first check if the key types are string
+        final Set keys = serverInfoMap.keySet();
+        if (keys.isEmpty()) {
+            return null;
+        }
+
+        final Object initialKey = keys.iterator().next();
+        if (initialKey instanceof String) {
```
jeddcuison: I think we can use this instead

    if (keys.contain(STATE_KEY)) {
        ...
    }

We won't have an invalid type error since contains uses the hash code to look for the key and I think the code would be easier to understand.

jeddcuison (commented):
Please see additional comments

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
```diff
@@ -69,18 +70,73 @@ public void receive(final Message jGroupMsg) {
...
+    private String getStringFromMessageMap(final Map serverInfoMap, final String key) {
+
+        // first check if the key types are string
+        final Set keys = serverInfoMap.keySet();
+        if (keys.isEmpty()) {
+            return null;
+        }
+
+        final Object initialKey = keys.iterator().next();
+        if (initialKey instanceof String) {
```
jeddcuison: Can we instead test for ReportingJmsMessageKey first ? e.g. 

    if (initialKey instanceof ReportingJmsMessageKey) {
        ...
       return ...
    }
    
    return (String) serverInfoMap.get(key);

If so we can skip the comment "assume...ReportingJmsMessageKey"

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
jmahmoodcc: Done

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
jmahmoodcc: No, it's used to get the NAME value from the message map as well.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
jmahmoodcc: Done

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
jmahmoodcc: Done - I don't know why I didn't do it for the key but I did it for the server map

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
jmahmoodcc: Done

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
jmahmoodcc: I can't though because I need to specifically know if the key is a String - if it is then I have to process it differently then if it's a ReportingJmsMessageKey object. We can't just check that the key is in the map, we have to test it's type too.

jeddcuison: Yes I understand but I was thinking the key "STATE" will have a different hash code to a key of type ReportingJmsMessageKey. I actually tried it. Anyways, we can also go with checking on the type but can we make it consistent with the other method e.g. check instance of ReportingJmsMessageKey instead of String ? Then we can ditch the "assume the key is...JvmState".

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
jeddcuison: Okay, it's just that it made me wonder why String in the method name when it returns a String already. It made me think is there something I'm missing. Then I realized the method is just getting a "String" value so getValFromMessageMap will remove the redundancy and remove the focus on the why "String". Anyways it's minor.

jeddcuison (commented):
Just one more comment and 1 minor - the minor you can choose to ignore

jwala-services/src/main/java/com/cerner/jwala/service/jvm/state/JvmStateReceiverAdapter.java
jmahmoodcc: We can't check for type of ReportingJmsMessageKey either. The main problem is that the old way sends the message map with the ReportingJmsMessageKey as a serialized object, but the ReportingJmsMessageKey only exists in the ctp-infrastructure libraries. So, jwala has no idea what this class is and we can't import it without adding a dependency to ctp-infrastructure which we definitely don't want.

It's unfortunate but we can only rely on the keys being Strings to determine which version of the message we're receiving. We really can't even do keys.contains(STATE_KEY) to check for a ReportingJmsMessageKey because we can't create a ReportingJmsMessageKey object.

jmahmoodcc: Ok, got it now. I'll change it use contains(). 

This will actually be more flexible if there are any future changes since jwala owns the tomcat-agent module now.

jeddcuison (approved):

PeterTeunissen (approved):

PR #102, opened 2017-05-11
Fix intermittent resource tab bug wherein add resource is not clickab…

jeddcuison:
This is a fix for the tests about external properties in the resource tab in wherein if the "Ext Properties" node is clicked the resource add button is momentarily disabled with a screen overlay (The gray thingy with the rotating gears) then executing a click on it makes Selenium throw an error stating that the "element is not clickable". The bug is intermittent and will only appear if the "Ext Properties" loading is delayed a little.

### DIFF
--- a/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/testsuite/configuration/resources/DeleteExternalProperty.java
+++ b/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/testsuite/configuration/resources/DeleteExternalProperty.java
@@ -16,7 +16,11 @@ public class DeleteExternalProperty extends JwalaTest {
     public void delExternalProperties() {
         clickTab("Configuration");
         clickTab("Resources");
-        driver.findElement(By.xpath("//li/span[text()='ext.properties']/preceding-sibling::input[@type='checkbox']")).click();
+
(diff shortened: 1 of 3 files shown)


jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/testsuite/configuration/resources/DeleteExternalProperty.java
```diff
@@ -16,7 +16,11 @@
     public void delExternalProperties() {
         clickTab("Configuration");
         clickTab("Resources");
-        driver.findElement(By.xpath("//li/span[text()='ext.properties']/preceding-sibling::input[@type='checkbox']")).click();
+
+        // If not coming from add ext property, this code makes sure ext properties is the currently selected
+        // Topology node
+        driver.findElement(By.xpath("//span[text()='Ext Properties']")).click();
```
akinny, line 22: This still has click, instead of waitClick, is this correct?

jeddcuison: Yes Arvindo. I only do a wait click when the previous action triggers the screen overlay with the gears.

jmahmoodcc (approved):

PR #104, opened 2017-05-12
Fix upload resource error message

jmahmoodcc:
When uploading a resource using a meta data file, if the meta data had a null entity the upload would fail with a generic error message.

Changed the REST layer to not catch the service exception so that the UI can display any error messages coming from the service layer.


LinHungWu (approved):

jwala-services/src/main/java/com/cerner/jwala/service/resource/impl/ResourceServiceImpl.java
```diff
@@ -168,6 +168,11 @@ public CreateResourceResponseWrapper createTemplate(final InputStream metaData,
             }
 
             if (StringUtil.isEmpty(targetName)) {
+                if (null == resourceTemplateMetaData.getEntity()) {
+                    String errMsg = "No entity data found in the meta data. Unable to continue with resource upload.";
```
akinny: Can you add some information about the resource so it will be easy tro debug

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/resource/impl/ResourceServiceRestImpl.java
```diff
@@ -62,42 +61,36 @@ public ResourceServiceRestImpl(final ResourceService resourceService) {
...
-            // TODO check for a max file size
-            List<Attachment> filteredAttachments = new ArrayList<>();
-            for (Attachment attachment : attachments) {
-                if (attachment.getDataHandler().getName() != null) {
-                    filteredAttachments.add(attachment);
-                }
+        // TODO check for a max file size
+        List<Attachment> filteredAttachments = new ArrayList<>();
+        for (Attachment attachment : attachments) {
+            if (attachment.getDataHandler().getName() != null) {
```
akinny: check for null for datahandler, could cause NPE

jwala-services/src/main/java/com/cerner/jwala/service/resource/impl/ResourceServiceImpl.java
jmahmoodcc: Done - added the deploy filename to the message

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/resource/impl/ResourceServiceRestImpl.java
jmahmoodcc: add NP check for dataHandler

akinny (approved):

PR #105, opened 2017-05-15
Replace hard coded location protocol "https" with the one derived fro…

jeddcuison:
The web server balance manager link is always set to use "https" since it was hard coded. This change replaces the hard coded https with a location protocol that was derived from the web server status path. If the status path is relative, the location protocol will be the location protocol of the current URL address.

### DIFF
--- a/jwala-webapp/src/main/webapp/resources/js/react/group-operations/WebServerControlPanelWidget.js
+++ b/jwala-webapp/src/main/webapp/resources/js/react/group-operations/WebServerControlPanelWidget.js
@@ -141,7 +141,8 @@ var WebServerControlPanelWidget = React.createClass({
         parseUrlProtocol: function(url) {
             try {
                 let protocol = url.split("/")[0].toLowerCase();
-                if (protocol.indexOf(WebServerControlPanelWidget.HTTP_PROTOCOL) > -1) {
+                if (protocol.indexOf(WebServerControlPanelWidget.HTTP_PROTOCOL) > -1 ||
+                    protocol.indexOf(WebServerControlPanelWidget.HTTPS_PROTOCOL) > -1) {
                     return protocol;
                 }
                 console.warn('URL "%s" does not have a location protocol. ' +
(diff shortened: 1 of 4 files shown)


jwala-webapp/src/main/webapp/resources/js/react/group-operations/WebServerControlPanelWidget.js
```diff
@@ -134,7 +137,23 @@ var WebServerControlPanelWidget = React.createClass({
     statics: {
         getReactId: function(dom) {
             return $(dom).attr("data-reactid");
-        }
+        },
+        parseUrlProtocol: function(url) {
+            try {
+                let protocol = url.split("/")[0].toLowerCase();
+                if (protocol.indexOf(WebServerControlPanelWidget.HTTP_PROTOCOL) > -1) {
```
jmahmoodcc: Since this is testing for "http:", isn't it possible that "https" won't be returned either if the location.protocol is "http" too?

jmahmoodcc (changes requested):
Just a question about how the protocol is derived

LinHungWu:
What is the default value if there is nothing to derived to?

jwala-webapp/src/main/webapp/resources/js/react/group-operations/WebServerControlPanelWidget.js
jeddcuison: Good catch! I changed the code.

jeddcuison:
Answer to @LinHungWu's question "What is the default value if there is nothing to derived to?"

As discussed with @akinny, the default is "http:"

jwala-webapp/src/main/webapp/resources/js/react/group-operations/WebServerControlPanelWidget.js
```diff
@@ -108,7 +108,10 @@ var WebServerControlPanelWidget = React.createClass({
     },
 
     onClickStatusLink: function(e) {
-        let statusUrl = "https://" + this.props.data.host + ":" + this.props.data.httpsPort + jwalaVars.loadBalancerStatusMount
+        let locationProtocol =  WebServerControlPanelWidget.parseUrlProtocol(this.props.data.statusPath.path);
+        let port = locationProtocol === WebServerControlPanelWidget.HTTP_PROTOCOL ? this.props.data.port :
```
akinny: shouldn't this.props.data.port be
this.props.data.httpPort?

akinny (approved):

PR #121, opened 2017-05-26
Fix war file generation for hostnames in different formats

akinny:
- Added lock on IP address at group level
- Optimize get host IP address call for jvm service

Before you contribute, please review these guidelines to help ensure a smooth process for everyone.

Thanks.

* Read [how to properly contribute to open source projects on Github][1].
* Fork the project.
* Use a feature branch.
* Write [good commit messages][2].
* Use the same coding conventions as the rest of the project.
* Commit locally and push to your fork until you are happy with your contribution.
* Make sure to add tests and verify all the tests are passing when merging upstream.
* Add an entry to the [Changelog][3] accordingly.
* Please add your name to the [CONTRIBUTORS.md][7] file. Adding your name to the [CONTRIBUTORS.md][7] file signifies agreement to all rights and reservations provided by the [License][4].
* [Squash related commits together][5].
* Open a [pull request][6].
* The pull request will be reviewed by the community and merged by the project committers.

[1]: http://gun.io/blog/how-to-github-fork-branch-and-pull-request
[2]: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[3]: ./CHANGELOG.md
[4]: ./LICENSE
[5]: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html
[6]: https://help.github.com/articles/using-pull-requests
[7]: ./CONTRIBUTORS.md

### DIFF
--- a/jwala-services/src/main/java/com/cerner/jwala/service/group/impl/GroupServiceImpl.java
+++ b/jwala-services/src/main/java/com/cerner/jwala/service/group/impl/GroupServiceImpl.java
@@ -1,5 +1,6 @@
 package com.cerner.jwala.service.group.impl;
 
+import com.cerner.jwala.common.JwalaUtils;
 import com.cerner.jwala.common.domain.model.app.Application;
 import com.cerner.jwala.common.domain.model.fault.FaultType;
 import com.cerner.jwala.common.domain.model.group.Group;
@@ -22,6 +23,7 @@ import com.cerner.jwala.common.rule.NameLengthRule;
 import com.cerner.jwala.common.rule.group.GroupNameRule;
 import com.cerner.jwala.persistence.service.ApplicationPersistenceService;
 import com.cerner.jwala.persistence.service.GroupPersistenceService;
+import com.cerner.jwala.service.binarydistribution.BinaryDistributionLockManager;
 import com.cerner.jwala.service.exception.GroupServiceException;
 import com.cerner.jwala.service.group.GroupService;
 import com.cerner.jwala.service.jvm.JvmService;
@@ -54,6 +56,9 @@ public class GroupServiceImpl implements GroupService {
 
     @Autowired
     private JvmService jvmService;
+
+    @Autowired
+    private BinaryDistributionLockManager binaryDistributionLockManager;
 
     public GroupServiceImpl(final GroupPersistenceService groupPersistenceService,
                             final ApplicationPersistenceService applicationPersistenceService,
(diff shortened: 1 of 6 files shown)


jwala-services/src/main/java/com/cerner/jwala/service/group/impl/GroupServiceImpl.java
```diff
@@ -411,7 +416,13 @@ protected CommandOutput executeDeployGroupAppTemplate(final String groupName, fi
                 .setGroupName(groupName)
                 .setWebAppName(application.getName())
                 .build();
-        return resourceService.generateAndDeployFile(resourceIdentifier, application.getName(), fileName, hostName);
+        String lockKey = groupName + application.getName() + fileName + JwalaUtils.getHostAddress(hostName);
```
jmahmoodcc, line 419: We need to do this for the group deploy of the JVM and Web Server resources are well. Unfortunately these methods are also contained in the REST layer so you would want to refactor those methods into the service layer as well.

jmahmoodcc (changes requested):
Add the lock to deploying the group JVM and Web Server resources. Requires some refactoring.

jwala-services/src/main/java/com/cerner/jwala/service/group/impl/GroupServiceImpl.java
akinny: This already done, where ever there was a host used in the key, it has been changed to IP address. Look at ApplicationServiceImpl.java

PeterTeunissen: Doesn't this change behavior? Seems that we did not use to lock before, but now we do?

PeterTeunissen (commented):
See comment.... If that's fine, then good to go.

jeddcuison (approved):

jwala-services/src/main/java/com/cerner/jwala/service/group/impl/GroupServiceImpl.java
jmahmoodcc: I think we didn't lock it before because we always assumed that the resource was only being deployed once per host. I was looking at the REST code, and the way it works is that we add the hostname for each JVM in the group to a list, and if the hostname is already in the list then we skip that deploy. So, as long as the hosts are in the same format only one would get deployed per host.

I'm thinking now, that with this change, the resource will be deployed multiple times for the same host if the hostname is specified for different JVMs using the short name, FQDN, and IP address.

Would it make more sense to populate the list of JVM hosts with the derived IP address and test against that instead?

I'm talking about this code in GroupServiceRestImpl.performGroupAppDeployToHosts (lines 769-779):

            List<String> deployedHosts = new ArrayList<>(jvms.size());
            for (final Jvm jvm : jvms) {
                final String hostName = jvm.getHostName();
                if (!deployedHosts.contains(hostName)) {
                    historyFacadeService.write(hostName, group, "Deploying application resource " + fileName, EventType.USER_ACTION_INFO, aUser.getUser().getId());
                    deployedHosts.add(hostName);
                    Future<Response> response = createFutureResponseForAppDeploy(groupName, fileName, appName, jvm, null);
                    if (response != null)
                        futureMap.put(hostName, response);
                }
            }

PR #122, opened 2017-05-26
Fix multiple binary uploads

jmahmoodcc:
When uploading media or a resource, check that no existing binaries already match the uploaded file. If an existing binary file matches then map the media to the currently existing one, and don't keep the newly uploaded binary file.


jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -101,6 +109,46 @@ public JpaMedia create(final Map<String, String> mediaDataMap, final Map<String,
...
+                        LOGGER.warn("Failed to delete uploaded file {}", uploadedFilePath, e);
+                    }
+                    return existingBinaryAbsPath;
+                }
+            }
+        }
+        return uploadedFilePath;
+    }
+
+    private String getCheckSum(final String fileAbsolutePath) {
```
akinny: Can you move this method to JwalaUtils or FileUtils?

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -101,6 +109,46 @@ public JpaMedia create(final Map<String, String> mediaDataMap, final Map<String,
                 format("{0} does not have any root directories! It may not be a valid media file.", filename));
     }
 
+    private String checkForExistingBinary(final String filename, final String uploadedFilePath) {
```
akinny: Can you move this method to JwalaUtils or FileUtils?

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
```diff
@@ -101,6 +106,37 @@ public JpaMedia create(final Map<String, String> mediaDataMap, final Map<String,
                 format("{0} does not have any root directories! It may not be a valid media file.", filename));
     }
 
+    private String checkForExistingBinary(String filename, String initialDestHash) {
+        List<String> binariesAbsPath = repositoryService.getBinariesByBasename(filename);
+        if (!binariesAbsPath.isEmpty()) {
```
akinny: Add binariesAbsPath!=null, just in case the directory doesn't have any files.

jwala-services/src/main/java/com/cerner/jwala/service/repository/AbstractRepositoryService.java
```diff
@@ -48,6 +54,17 @@ public void delete(final String filename) {
         }
     }
 
+    @Override
+    public List<String> getBinariesByBasename(String filename) {
+        File binariesDir = new File(getRepositoryPath().toAbsolutePath().normalize().toString());
+        File[] foundFiles = binariesDir.listFiles((dir, name) -> name.contains(filename));
```
akinny: Can you do a case insensitive check?

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
jmahmoodcc: moved to FileUtility

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
jmahmoodcc: This method relies on the RepositoryService. Are we ok to add that dependency to JwalaUtils?

jwala-services/src/main/java/com/cerner/jwala/service/repository/AbstractRepositoryService.java
jmahmoodcc: changed it to compare against lower case file names

PeterTeunissen (approved):

jwala-services/src/main/java/com/cerner/jwala/service/media/impl/MediaServiceImpl.java
jeddcuison: My opinion here is that if it's only used once then a private method in the class is okay. In addition I heard that it's a good rule of thumb not to assume that a certain method would be reused in other places as well. Until that happens then that's the time to relocate/refactor it.

jeddcuison (approved):

PR #124, opened 2017-05-31
Make JwalaChromeTestSuite generic

jeddcuison:
By generic it means that the test suite can be executed against any browser that selenium supports

### DIFF
--- a/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/JwalaTestSuite.java
+++ b/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/JwalaTestSuite.java
@@ -34,13 +34,13 @@ import java.util.concurrent.TimeUnit;
  *
  * 1. web driver
  *
- *    Chrome:    -DwebDriver=org.openqa.selenium.chrome.ChromeDriver
- *    IE:        -DwebDriver=org.openqa.selenium.ie.InternetExplorerDriver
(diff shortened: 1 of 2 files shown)


jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/JwalaTestSuite.java
```diff
@@ -36,24 +51,22 @@
         AppDeleteTest.class, JvmOperationsPageDeleteTest.class, WebServerOperationsPageDelete.class, WebServerCreateTest.class,
         WebServerDeleteTest.class, JvmCreateTest.class, JvmDeleteTest.class, GroupDeleteTest.class, BalancerManagerTest.class,
         LogoutTest.class})
-public class JwalaChromeTestSuite extends TestSuite {
+public class JwalaTestSuite extends TestSuite {
 
-    private static final String WEBDRIVER_CHROME_DRIVER = "webdriver.chrome.driver";
     private static final String ELEMENT_SEARCH_RENDER_WAIT_TIME = "element.search.render.wait.time";
+    public static final String WEB_DRIVER = "webDriver";
```
jmahmoodcc: Just a very minor thing: should this not be camel case? It looks like it's read as a System Property, but is it ever read from the properties file too?

jmahmoodcc (commented):
Quick question about the web driver key name. That's all.

LinHungWu (approved):

PR #128, opened 2017-06-09
Mask Windows service password in the logs

jeddcuison:
The masking is done via Log4j layout


jwala-services/src/test/java/com/cerner/jwala/service/custom/logging/WinSvcPasswordMaskingLayoutTest.java
```diff
@@ -0,0 +1,103 @@
...
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Unit tests for {@link WinSvcPasswordMaskingLayout}
+ * Created by Jedd Cuison on 6/7/2017
+ */
+public class WinSvcPasswordMaskingLayoutTest {
+
+    private static final String LOG_MSG_WITH_PASSWORD_1 = "Executing command \"~/.jwala/CTO-N9SF-LTST-HEALTH-CHECK-4.0-" +
+            "USMLVV1CDS0049-2/install-service.sh CTO-N9SF-LTST-HEALTH-CHECK-4.0-USMLVV1CDS0049-2 D:/ctp/app/instances " +
```
akinny: Can you take out all our server names

jwala-services/src/test/java/com/cerner/jwala/service/custom/logging/WinSvcPasswordMaskingLayoutTest.java
```diff
@@ -0,0 +1,103 @@
...
+            "apache-tomcat-7.0.55 \"jedi\" the-password \"";
+    private static final String LOG_MSG_WITH_PASSWORD_2 = "D:\\cygwin64\\home\\N9SFGLabTomcatAdmin>set svc_password=the-password ";
```
akinny: remove internal user name

jwala-services/src/main/java/com/cerner/jwala/service/custom/logging/WinSvcPasswordMaskingLayout.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        final String msg = event.getRenderedMessage();
+        String maskedMsg = StringUtils.EMPTY;
+
+        // This layout is specific for svc_password masking therefore the code is straight forward and not generic
+        if (StringUtils.indexOf(msg, "install-service.sh") > 0) {
+            int index = msg.lastIndexOf(' ', msg.lastIndexOf(' ') - 1);
+            maskedMsg = msg.substring(0, index) + " ******** \"";
+        } else if (StringUtils.contains(msg, "set svc_password=")) {
```
akinny: Can you move the hard coded strings as constants, so it will be easy to change if required?

akinny (commented):
Do we have to change jwala/jwala-tomcat/src/main/resources/data/conf/log4j.xml for applying this override?

jwala-services/src/main/java/com/cerner/jwala/service/custom/logging/WinSvcPasswordMaskingLayout.java
jeddcuison: done

jeddcuison:
@akinny, I updated the log4j xmls as well as per your comment

jeddcuison:
Please note that I forced my latest commit to overwrite codes that have sensitive info

jwala-services/src/main/java/com/cerner/jwala/service/custom/logging/WinSvcPasswordMaskingLayout.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.*;
+
+/**
+ * Masks the Windows Service Password (svc_password)
+ * Created by Jedd Cuison on 6/6/2017
+ */
+public class WinSvcPasswordMaskingLayout extends PatternLayout {
+
+    private static final String INSTALL_SERVICE_SH = "install-service.sh";
+    private static final String SET_SVC_PASSWORD = "set svc_password=";
```
jmahmoodcc: Is it possible to create the above 2 constants as parameters in the log4j.xml?

I'm just thinking that these searches are very specific to the current scripts, and in case it changes in the future it would be easier to just update the log4j.xml then have to rebuild the whole project.

jmahmoodcc (commented):
Question about parameterizing the search expressions

akinny (approved):

jwala-services/src/main/java/com/cerner/jwala/service/custom/logging/WinSvcPasswordMaskingLayout.java
jeddcuison: Working on it...

PeterTeunissen:
Instead of searching for specific keywords and scrub those, I think the scrubber should have a static list of  strings that should be scrubbed. Then, when logging a line, replace all the occurrences of the strings in the list with '****'.  You populate the list of static strings whenever you call the password decrypt routine (so with the clear text pw). This way nobody can forget to change the scrubber when you change the shell scripts or other keywords....

public class Scrubber {

    private static HashMap<String, String> toRemove = new HashMap<String, String>();

    public static void addKey(String key) {
        String mask = "";
        for (int i = 0; i < key.length(); i++) {
            mask += "*";
        }
        toRemove.put(key, mask);
    }

    public static String scrubMe(String raw) {
        String clean = raw;
        for (String key : toRemove.keySet()) {
            clean = clean.replaceAll(key, toRemove.get(key));
        }
        return clean;
    }

    public static void main(String[] args) {
        System.out.println(scrubMe("My Password is: secret"));
        Scrubber.addKey("secret");
        System.out.println(scrubMe("My Password is: secret"));
    }
}

Output:
My Password is: secret
My Password is: ******

PeterTeunissen (commented):
See my comment... just my $0.02

akinny:
I like it Peter!! Just make the put in a synchronized block

jeddcuison:
Thanks @PeterTeunissen, we implemented your solution

jwala-services/src/main/java/com/cerner/jwala/service/custom/logging/impl/JvmWinSvcAcctPasswordScrubber.java
```diff
@@ -0,0 +1,95 @@
...
+    private final Set<String> toRemove = new HashSet<>();
+    private DecryptPassword decryptor;
+    private final String [] includedMethodsArray = {"getExecRemoteCommandReturnInfo", "runExecCommand"};
+    private final Set<String> includedMethods = new HashSet<>(Arrays.asList(includedMethodsArray));
+
+    public JvmWinSvcAcctPasswordScrubber(final JvmPersistenceService jvmPersistenceService, final DecryptPassword decryptor) {
+        this.decryptor = decryptor;
+
+        // Populate the map of items to remove in the logs
+        final List<Jvm> jvms = jvmPersistenceService.getJvms();
```
akinny: When is this initialized? What happens to newly added JVM's with service accounts

jmahmoodcc: What if the map was populated whenever a JVM is added, updated, or removed? You'd probably have to synchronize access to it as well.

jeddcuison:
This was closed in lieu of https://github.com/cerner/jwala/pull/137

PR #130, opened 2017-06-12
SOARCOM-2962 From OPL: Switching TOC's JNDIRealm to use ldaps (secure…

LinHungWu:
…) does not work

- Add h2_keystore.jks under data/security/jks dir

- The new keystore for h2 should works for both windows and linux machine.

After the fix the jwala should perform following:
1. JWALA should use ldaps to login with domain account
2. JWALA H2 should still works with https://xxx:8084


akinny (commented):
@LinHungWu  can you add some details to the trust store. Like country, state, city, org etc?

jmahmoodcc (approved):

PR #137, opened 2017-06-20
Mask password (if there are any) from Jsch output and its debug log

jeddcuison:
Mask Jsch output and the command string before logging to prevent clear text password from filtering down the classes that uses the Jsch service and from showing up in the logs.


akinny (commented):
Can you add a test with a String similar to returned by jsch then, run the scrubber with and without the password?

jeddcuison:
@akinny, I followed your suggestion please check. Thanks!

PR #138, opened 2017-06-21
Remove the CopyWriteArraySet in the interface

jeddcuison:
Having an implementation variable in the interface is bad practice (my bad) since it violates encapsulation


jeddcuison:
This branch should NOT HAVE any conflicts

PR #139, opened 2017-06-21
Remove CopyOnWriteArraySet in the interface and move it into the impl…

jeddcuison:
Having an implementation variable in the interface is bad practice since it violates encapsulation


akinny (approved):

jmahmoodcc (approved):
Looks good to me

PR #141, opened 2017-06-29
Rewrite selenium tests to use cucumber

jeddcuison:


jeddcuison:
Please see https://github.com/cerner/jwala/pull/142

PR #142, opened 2017-06-29
Cucumber Tests

jeddcuison:
Write comprehensive Cucumber-Selenium tests to help lower the time it takes to do regression testing. In addition the other goal is to provide UI code coverage and to provide readable scenarios through Cucumber.


jmahmoodcc (approved):
Looks like a really good start. How long does it take to run all the tests we have so far?

PR #143, opened 2017-06-30
Integrate Apache HTTPD 2.4.26 into stp-toc and stp-healthcheck cookbook

LinHungWu:
Make sure jwala drain feature can handle multiple version of apache httpd such as 2.4.10, 2.4.20 and 2.4.26. 

*. Update the balancer manager class to parse any of known apache httpd html

### DIFF
--- a/jwala-services/src/main/java/com/cerner/jwala/service/balancermanager/impl/BalancerManagerHtmlParser.java
+++ b/jwala-services/src/main/java/com/cerner/jwala/service/balancermanager/impl/BalancerManagerHtmlParser.java
@@ -14,20 +14,21 @@ public class BalancerManagerHtmlParser {
     private static final Logger LOGGER = LoggerFactory.getLogger(BalancerManagerHtmlParser.class);
 
     public String getUrlPath(final String host, final int httpsPort) {
-        return "https://" + host + ":" +  httpsPort + "/balancer-manager";
+        return "https://" + host + ":" + httpsPort + "/balancer-manager";
     }
 
     public String getWorkerUrlPath(final String host, final int httpsPort, final String balancerName, final String nonce, final String workerName) {
-        return "https://" + host + ":" +  httpsPort + "/balancer-manager" + "?b=" + balancerName + "&nonce=" + nonce + "&w=" + workerName;
+        return "https://" + host + ":" + httpsPort + "/balancer-manager" + "?b=" + balancerName + "&nonce=" + nonce + "&w=" + workerName;
     }
 
     public Map<String, String> findBalancers(final String content) {
         LOGGER.info("Entering findBalancers:");
         Map<String, String> balancers = new HashMap<>();
-        final String foundStringPrefix = "<h3>LoadBalancer Status for <a href=\"/balancer-manager\\?b=";
-        final String foundStringPost = "\\&nonce=";
-        final String foundStringPostNonce = "\">balancer://";
-        final String matchPattern = foundStringPrefix + ".*" + foundStringPost + ".*";
+        final String foundStringPrefix = "<h3>LoadBalancer Status for <a href=..balancer-manager\\?b=";
+        final String foundStringPost = "nonce=";
+        final String foundStringPostNonce = ">balancer";
+        final String matchStringBalancerNamePrefix = "?b=";
+        final String matchPattern = foundStringPrefix + ".*." + foundStringPost + ".*";
         String balancerName;
         String nonce;
         Pattern pattern = Pattern.compile(matchPattern);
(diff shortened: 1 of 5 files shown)


jwala-services/src/main/java/com/cerner/jwala/service/balancermanager/impl/BalancerManagerHtmlParser.java
```diff
@@ -14,34 +14,38 @@
...
-            nonce = matchString.substring(
-                    matchString.indexOf(foundStringPost.replace("\\", "")) + foundStringPost.replace("\\", "").length(),
-                    matchString.indexOf(foundStringPostNonce));
+            balancerName = matchString.substring(matchString.indexOf(matchStringBalancerNamePrefix) +
+                    matchStringBalancerNamePrefix.length(), matchString.indexOf("&"));
+            nonce = matchString.substring(matchString.indexOf(foundStringPost) + foundStringPost.length(),
+                    matchString.indexOf(foundStringPostNonce) - 1);
             balancers.put(balancerName, nonce);
         }
+        LOGGER.info("balancers.size: " + balancers.size());
```
jmahmoodcc, line 45: These look like logging statements for debugging. Either clean them up or make the debug level logging.

jmahmoodcc (changes requested):
Looks good, just clean up the one logging section.

akinny (approved):
Change logging from info to debug

jwala-services/src/main/java/com/cerner/jwala/service/balancermanager/impl/BalancerManagerHtmlParser.java
LinHungWu: This is logger with info with purpose in jwala.log so we can easily to see the balancer manager setup

LinHungWu:
Please review again with 2 versions of httpd.conf in CI process.

jwala-services/src/main/java/com/cerner/jwala/service/balancermanager/impl/BalancerManagerHtmlParser.java
```diff
@@ -14,34 +14,38 @@
...
+        for (Map.Entry entry : balancers.entrySet()) {
+            LOGGER.info("balancerName: " + entry.getKey() + ", nonce: " + entry.getValue());
```
jmahmoodcc, line 47: Can we move this logging statement to line 43 inside the while loop? We shouldn't have logic specifically for logging, especially since we capture all this data in the while loop while building the balancers set.

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/resource/impl/ResourceServiceRestImpl.java
```diff
@@ -132,17 +132,29 @@ public Response createResource(final String deployFilename, final CreateResource
...
-                    "Invalid number of attachments! " + CREATE_RESOURCE_ATTACHMENT_SIZE + " attachments is expected by the service."));
+        if (attachments == null || !isExternalProperty && attachments.size() == 0 ) {
+            LOGGER.error("Expected non-empty attachments. Returning without creating Resource.");
+            return Response.status(Response.Status.BAD_REQUEST).build();
+        }
+
+        if (attachments != null){
+            LOGGER.debug("attachments.size() " + attachments.size());
+            for (final Attachment attachment : attachments) {
+                LOGGER.debug("attachment.getDataHandler().getName(): " + attachment.getDataHandler().getName() +
```
jmahmoodcc: Same thing here with the logging: please move these logging statements into the for-loop below.

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/resource/impl/ResourceServiceRestImpl.java
```diff
@@ -132,17 +132,29 @@ public Response createResource(final String deployFilename, final CreateResource
...
-        if (attachments == null || !isExternalProperty && attachments.size() != CREATE_RESOURCE_ATTACHMENT_SIZE) {
-            return ResponseBuilder.notOk(Response.Status.INTERNAL_SERVER_ERROR, new FaultCodeException(
-                    FaultType.INVALID_NUMBER_OF_ATTACHMENTS,
-                    "Invalid number of attachments! " + CREATE_RESOURCE_ATTACHMENT_SIZE + " attachments is expected by the service."));
+        if (attachments == null || !isExternalProperty && attachments.size() == 0 ) {
+            LOGGER.error("Expected non-empty attachments. Returning without creating Resource.");
+            return Response.status(Response.Status.BAD_REQUEST).build();
+        }
+
+        if (attachments != null){
```
jmahmoodcc: Should the method exit early if there are no attachments provided? Would it make more sense to throw an exception here if there are no attachments?

jmahmoodcc (changes requested):
See my comments about logging and check for no attachments in the ResourceService REST API

jwala-services/src/main/java/com/cerner/jwala/service/balancermanager/impl/BalancerManagerHtmlParser.java
LinHungWu: If we move inside line 43, then if no find , i do not get the logging for size=0. so it is not easy for debugging.

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/resource/impl/ResourceServiceRestImpl.java
LinHungWu: Create RESOURCE_NO_ATTACHMENT_EXCEPTION("AEM78", "ResourceNoAttachmentException");

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/resource/impl/ResourceServiceRestImpl.java
LinHungWu: done

PR #144, opened 2017-07-11
Remove remote.jwala.webapps.dir

jmahmoodcc:
Remove all references to the vars.properties property remote.jwala.webapps.dir to remove use of an absolute path in order to support multiple operating systems from a single jwala instance.
 - Delete all test properties and application properties references
 - Create a new attribute in the application model to store the value
 - Read the value from the meta data of the war of the application and keep the application attribute updated when the meta data changes
 - Update the postman tests to no longer use the property


jwala-common/src/main/java/com/cerner/jwala/common/domain/model/webserver/WebServer.java
```diff
@@ -206,4 +208,12 @@ public String toString() {
                 ", state=" + state +
                 '}';
     }
+
+    private String getGroupNames(Collection<Group> groups) {
+        StringBuilder groupsNames = new StringBuilder();
+        for (Group group : groups) {
+            groupsNames.append(group.getName()).append(" ");
```
akinny: Can you add a comma separator, if the group name has a space in it you cannot tokenize?

jmahmoodcc: done

akinny (approved):

jwala-services/src/main/java/com/cerner/jwala/service/app/impl/ApplicationServiceImpl.java
```diff
@@ -403,6 +403,23 @@ private void copyAndExecuteCommand(Application application, Set<String> hostName
...
+                .setResourceName(appWarName)
+                .setWebAppName(application.getName())
+                .setGroupName(application.getGroup().getName())
+                .build();
+        String metaData = resourceService.getResourceContent(appWarResourceId).getMetaData();
+        try {
+            return resourceService.getTokenizedMetaData(appWarName, application, metaData).getDeployPath();
+        } catch (IOException e) {
+            String messageErr = MessageFormat.format("Failed to generate the war meta data for {0}", application);
+            LOGGER.error(messageErr);
```
jeddcuison: I think we need to include the original exception with the logs and when throwing it to have the original cause of the error as a stack trace.

    LOGGER.error(messageErr, e);
    throw new ApplicationServiceException(messageErr, e);

Sonar also flags it when we don't include the original exception:

http://[SonarQube Server]:9000/component/index?id=workspace%3Ajwala-services%3Asrc%2Fmain%2Fjava%2Fcom%2Fcerner%2Fjwala%2Fservice%2Fgroup%2Fimpl%2FGroupServiceImpl.java&line=491

jwala-services/src/main/java/com/cerner/jwala/service/resource/impl/ResourceServiceImpl.java
```diff
@@ -1003,6 +1023,37 @@ private void doUnpack(final String hostName, final String destPath) {
         return resourceFileString;
     }
 
+    private Application setApplicationWarDeployPath(Application application) {
```
jeddcuison: The method does a lot of things...maybe we need to put java docs or maybe we can do something like this ?

    updatedApp = given(application).getWarResourceContent().getMetaData().tokenize().updateWarPath();

https://en.wikipedia.org/wiki/Fluent_interface

jeddcuison (commented):
Please see code comments.

jeddcuison:
Just to confirm, chef will handle upgrade scenarios ?

jwala-services/src/main/java/com/cerner/jwala/service/app/impl/ApplicationServiceImpl.java
jmahmoodcc: done

jwala-services/src/main/java/com/cerner/jwala/service/resource/impl/ResourceServiceImpl.java
jmahmoodcc: I like the idea of the fluent interface, but in the interest of time I just extracted some private methods and added a java doc. In the future I will definitely think more "fluently" :)

jmahmoodcc:
Yes, the chef cookbooks handle the upgrade scenarios

jwala-services/src/main/java/com/cerner/jwala/service/resource/impl/ResourceServiceImpl.java
jeddcuison: Ok :)

jeddcuison (approved):

PR #145, opened 2017-07-11
Set jgroups bind address java option

jeddcuison:
This solves the problem wherein the internal ip address does not work when there is no external network connected.


jwala-services/src/main/java/com/cerner/jwala/service/initializer/JGroupsClusterInitializer.java
```diff
@@ -61,6 +57,13 @@ public void afterPropertiesSet() throws Exception {
                 }
                 coordinatorIP = getIPAddressFromHostname(jgroupsCoordinatorHostname);
             }
+
+            System.setProperty(JAVA_NET_PREFER_IPV4_STACK, jgroupsJavaNetPreferIPv4Stack);
+            System.setProperty(JGROUPS_BIND_ADDR, coordinatorIP.getIpAddress().getHostAddress());
```
akinny: Can you add null check for getIpAddress to avoid NPE? I am not sure if this could be null, in some instances of multiple IP's,  it will not harm to add a null check.

jmahmoodcc (approved):
Approve with making Arvindo's requested change for checking for NPE

jmahmoodcc (approved):

akinny (approved):

PR #149, opened 2017-07-13
create jvm error when application name is not passed from rest call

rahulsayini:
Now application name is retrieved with group selected by user, On jvm creation


jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
```diff
@@ -68,12 +68,14 @@
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
 import org.springframework.test.context.support.AnnotationConfigContextLoader;
+import ucar.nc2.dataset.conv.COARDSConvention;
```
akinny: Is this class used? Can you remove it

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -218,7 +218,8 @@ public void createDefaultTemplates(final String jvmName, Group parentGroup) {
             try {
                 ResourceTemplateMetaData metaData = resourceService.getMetaData(metaDataStr);
                 if (metaData.getEntity().getDeployToJvms()) {
-                    final String template = resourceService.getAppTemplate(groupName, metaData.getEntity().getTarget(),
+                    String applicationName = applicationService.findApplications(groupPersistenceService.getGroup(groupName).getId()).iterator().next().getName();
```
akinny: You are assigning the first app, what if the group has multiple apps, will this work?

jwala-persistence/src/main/java/com/cerner/jwala/persistence/jpa/domain/resource/config/template/JpaGroupAppConfigTemplate.java
```diff
@@ -20,6 +20,9 @@
         @NamedQuery(name = JpaGroupAppConfigTemplate.GET_GROUP_APP_TEMPLATE_META_DATA,
                 query = "SELECT t.metaData FROM JpaGroupAppConfigTemplate t where LOWER(t.grp.name) = LOWER(:grpName)" +
                         " and t.templateName = :templateName"),
+        @NamedQuery(name = JpaGroupAppConfigTemplate.GET_GROUP_APP_TEMPLATE_META_DATA_WITH_APP,
+                query = "SELECT t.metaData FROM JpaGroupAppConfigTemplate t where LOWER(t.grp.name) = LOWER(:grpName)" +
+                        " and t.templateName = :templateName" + " and t.app.name = :appName"),
```
jeddcuison: Is this case insensitive ? 

    " and t.templateName = :templateName" + " and t.app.name = :appName"

I see that LOWER was used in group name, I would assume we also need to use it on the templateName and appName.

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
```diff
@@ -142,6 +140,40 @@ public void testCreateValidate() {
...
+        when(Config.mockGroupPersistenceService.getGroupAppsResourceTemplateNames(anyString())).thenReturn(templateNames);
+        when(Config.mockApplicationService.findApplications(any())).thenReturn(applications);
+        when(mockMetaData.getEntity()).thenReturn(mockEntity);
+        when(mockEntity.getDeployToJvms()).thenReturn(false);
+
+
+        jvmService.createDefaultTemplates("TestJvm", group);
+
+        System.clearProperty(ApplicationProperties.PROPERTIES_ROOT_PATH);
+    }
```
jeddcuison: We should "explicitly" test functionality here. It's tempting to just write test that does not throw any error and that would be the criteria that the test passes but if you think about it are we really testing the functionality not if it just runs without any errors so in lieu of that can we assert anything here ? If not can we verify something here ?

jwala-persistence/src/main/java/com/cerner/jwala/persistence/jpa/domain/resource/config/template/JpaGroupAppConfigTemplate.java
```diff
@@ -20,6 +20,9 @@
         @NamedQuery(name = JpaGroupAppConfigTemplate.GET_GROUP_APP_TEMPLATE_META_DATA,
```
jmahmoodcc, line 20: Remove this query and replace instances where this is used with the new query. We have to remove this query because it violates the model constraints now.

jwala-persistence/src/main/java/com/cerner/jwala/persistence/jpa/service/impl/GroupCrudServiceImpl.java
```diff
@@ -260,6 +260,20 @@ public void updateGroupAppResourceMetaData(String groupName, String webAppName,
...
+        try {
+            return (String) q.getSingleResult();
+        } catch (RuntimeException re) {
+            LOGGER.error("Error getting group app resource meta data for resource {} in group {}", templateName, groupName, re);
+            throw new NonRetrievableResourceTemplateContentException(groupName, templateName, re);
+        }
+    }
+
+    @Override
     public String getGroupAppResourceTemplateMetaData(String groupName, String resourceTemplateName) {
```
jmahmoodcc: Delete this method and replace it with the new method

jmahmoodcc (changes requested):
Remove the old method that didn't use the application name

jeddcuison (commented):
Since this is opensource can we put a short description on what the error is instead of just saying "Create JVM error" ? I had to look for the JIRA to understand what this pull request is trying to fix, people in the opensource community doesn't have access to our JIRA.

jeddcuison:
I remember that the main reason why there was an error creating the JVM is because the query that returns the template which is to be assigned to the JVM is not returning anything. The reason is because the web app name was taken from the metadata entity target which is null for non legacy resources. I think the solutions is just to get the application names and use them to get the template instead of meta data entity target. 

    final List<Application> appList = applicationService.findApplications(parentGroup.getId());

        // get the group App templates
        templateNames = groupPersistenceService.getGroupAppsResourceTemplateNames(groupName);
        for (String templateName : templateNames) {
            String metaDataStr = groupPersistenceService.getGroupAppResourceTemplateMetaData(groupName, templateName);
            try {
                ResourceTemplateMetaData metaData = resourceService.getMetaData(metaDataStr);
                if (metaData.getEntity().getDeployToJvms()) {
                    for (final Application app : appList) {
                        final String template = resourceService.getAppTemplate(groupName, app.getName(), templateName);
                        final ResourceIdentifier resourceIdentifier = new ResourceIdentifier.Builder()
                                .setResourceName(metaData.getTemplateName()).setJvmName(jvmName)
                                .setWebAppName(app.getName()).build();
                        resourceService.createResource(resourceIdentifier, metaData, new ByteArrayInputStream(template.getBytes(StandardCharsets.UTF_8)));
                    }
                }
            } catch (IOException e) {
                LOGGER.error("Failed to map meta data while creating JVM for template {} in group {}",
                        templateName, groupName, e);
                throw new InternalErrorException(FaultType.BAD_STREAM, "Failed to map data for template " +
                        templateName + " in group " + groupName, e);
            }
        }

jeddcuison (changes requested):
I check again your code and it exposed another problem which is related to group web app preview. The bug is that if the web app is not generated by chef e.g. HCT data set, the web app name in the meta data will be null thus preview will not work (this happens in the old code). So we can go forward with your solution to fix the preview problem since it fixes 2 problems at the same time albeit the changes are not minimal. So the problem now is that the changes you made in the UI does not have the web application name. You should pass the web application name so that preview will work. After doing so, please test the solution against CHEF generated HCT data that has the legacy metadata content.

jwala-persistence/src/main/java/com/cerner/jwala/persistence/jpa/domain/resource/config/template/JpaGroupAppConfigTemplate.java
```diff
@@ -49,7 +49,7 @@
 public class JpaGroupAppConfigTemplate extends ConfigTemplate {
     public static final String GET_GROUP_APP_TEMPLATE_RESOURCE_NAMES = "getGroupAppTemplateResourceNames";
     public static final String GET_GROUP_APP_TEMPLATE_CONTENT = "getGroupAppTemplateContent";
-    public static final String GET_GROUP_APP_TEMPLATE_META_DATA = "getGroupAppTemplateMetaData";
+    public static final String GET_GROUP_APP_TEMPLATE_META_DATA_WITH_APP = "getGroupAppTemplateMetaDataWithApp";
```
jeddcuison: Please change the name, since the name tells me that you are retrieving the meta data WITH the application. I think what you meant was just get the meta data. GET_GROUP_APP_TEMPLATE_META_DATA is fine since you deleted the old one. Parameters should be seldom used in the name.

jwala-persistence/src/main/java/com/cerner/jwala/persistence/jpa/service/impl/GroupCrudServiceImpl.java
```diff
@@ -260,15 +260,16 @@ public void updateGroupAppResourceMetaData(String groupName, String webAppName,
     }
 
     @Override
-    public String getGroupAppResourceTemplateMetaData(String groupName, String resourceTemplateName) {
-        final Query q = entityManager.createNamedQuery(JpaGroupAppConfigTemplate.GET_GROUP_APP_TEMPLATE_META_DATA);
+    public String getGroupAppResourceTemplateMetaDataWithAppname(String groupName, String templateName, String appName) {
```
jeddcuison: Change to getGroupAppResourceTemplateMetaData

jwala-persistence/src/main/java/com/cerner/jwala/persistence/service/GroupPersistenceService.java
```diff
@@ -118,4 +116,6 @@
     String updateGroupAppResourceMetaData(String groupName, String webAppName, String resourceName, String metaData);
 
     List<JpaGroup> findGroups(List<Long> idList);
+
+    String getGroupAppResourceTemplateMetaDataWithAppName(String groupName, String templateName, String appName);
```
jeddcuison: Change to getGroupAppResourceTemplateMetaData

jwala-persistence/src/main/java/com/cerner/jwala/persistence/service/impl/JpaGroupPersistenceServiceImpl.java
```diff
@@ -314,4 +310,9 @@ public boolean checkGroupWebServerResourceFileName(String groupName, String file
         q.setParameter(JpaGroup.QUERY_PARAM_ID_LIST, idList);
         return q.getResultList();
     }
+
+    @Override
+    public String getGroupAppResourceTemplateMetaDataWithAppName(String groupName, String templateName, String appName) {
```
jeddcuison: Change to getGroupAppResourceTemplateMetaData

jeddcuison (changes requested):
Fix preview is not working due to UI and backend changes

akinny (approved):

jwala-persistence/src/main/java/com/cerner/jwala/persistence/jpa/domain/resource/config/template/JpaGroupAppConfigTemplate.java
```diff
@@ -19,7 +19,7 @@
                         "LOWER(:grpName) and LOWER(t.app.name) = LOWER(:appName) and t.templateName = :templateName"),
         @NamedQuery(name = JpaGroupAppConfigTemplate.GET_GROUP_APP_TEMPLATE_META_DATA,
                 query = "SELECT t.metaData FROM JpaGroupAppConfigTemplate t where LOWER(t.grp.name) = LOWER(:grpName)" +
-                        " and t.templateName = :templateName"),
+                        " and t.templateName = :templateName" + " and t.app.name = :appName"),
```
jeddcuison: We should make templateName and appName comparison case insensitive

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -211,27 +211,30 @@ public void createDefaultTemplates(final String jvmName, Group parentGroup) {
...
+            templateNames = groupPersistenceService.getGroupAppsResourceTemplateNames(groupName, app.getName());
+            for (String templateName : templateNames) {
+                String metaDataStr = groupPersistenceService.getGroupAppResourceTemplateMetaData(groupName, templateName, app.getName());
+                try {
+                    ResourceTemplateMetaData metaData = resourceService.getMetaData(metaDataStr);
+                    if (metaData.getEntity().getDeployToJvms()) {
+                        final String template = resourceService.getAppTemplate(groupName, app.getName(),
+                                templateName);
+                        final ResourceIdentifier resourceIdentifier = new ResourceIdentifier.Builder()
+                                .setResourceName(metaData.getTemplateName()).setJvmName(jvmName)
```
jeddcuison: Can you just you the templateName so it's consistent with 

    getGroupAppResourceTemplateMetaData(groupName, templateName, ...)

jeddcuison (changes requested):
I have 2 more comments, the major one is - "We should make templateName and appName comparison case insensitive"
Once  the said comments are addressed I'll approve this pull request and merge the code

PR #172, opened 2017-10-19
Fix ui integ test ie

jeddcuison:
There were additional failing tests in IE that I had to fix

**The test for the unzip story is failing. The expectation of the test is that the deployment will fail due to unzip errors but it doesn't, the war is deployed successfully for some reason. The scenario was temporarily disabled so I can run the tests in the server (I'm checking whether the tests are running reliably in IE) that @ManickPrakash gave me while fixing the aforementioned problem in my local.**


jeddcuison:
Can you bring back the source manage control in ui-integration-tests-ie job ? Thanks

jeddcuison:
I was comparing the test results for Chrome and IE and I noticed that the External Properties tests are not in the IE report. Was the Selenium Grid not able to fix those tests ? Or they were removed before Selenium Grid and we just forgot to put them in ? Or you optimized it and placed it somewhere else ?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/configuration/ManageWsResourceTest.java
```diff
@@ -1,14 +0,0 @@
-package com.cerner.jwala.ui.selenium.configuration;
```
jeddcuison, line 1: Why was this removed ?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/operation/DeleteInOperationsTest.java
```diff
@@ -1,14 +0,0 @@
-package com.cerner.jwala.ui.selenium.operation;
```
jeddcuison, line 1: Why was this removed ?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/operation/StatusTest.java
```diff
@@ -1,14 +0,0 @@
-package com.cerner.jwala.ui.selenium.operation;
```
jeddcuison, line 1: We don't need this anymore ?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HandleResourceRunSteps.java
```diff
@@ -144,7 +144,6 @@ public void selectResourceCheckBox(String resourceFile) {
 
     @Then("^I confirm delete a resource popup$")
     public void deleteResource() {
-        jwalaUi.isElementExists(By.xpath("//[contains(text(),'Are you sure you want to delete the selected resource template(s) ?')]"));
```
jeddcuison, line 132: This was to confirm that the popup displayed is the delete resource confirmation popup. Why remove it ?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HotDeployRunSteps.java
```diff
@@ -76,14 +76,11 @@ public void verifyErrorMessageIndividualApp(String group, String file, String jv
 
     @Then("^I confirm webapp generate error popup in operations for jvm \"(.*)\"$")
     public void verifyAppOperationsDeployError(String jvmName) {
-        jwalaUi.isElementExists(By.xpath("//div[contains(text(),'Make sure the following JVMs are completely stopped before deploying.')]"));
```
jeddcuison: This is to confirm that we are seeing the correct popup before clicking okay. Why remove it ?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HotDeployRunSteps.java
```diff
@@ -76,14 +76,11 @@ public void verifyErrorMessageIndividualApp(String group, String file, String jv
...
-        jwalaUi.isElementExists(By.xpath("//div[text()='" + jvmName + "'"));
         clickOkButton();
     }
 
     @Then("^I confirm webapp \"([^\"]*)\" is successfully deployed in Operations page popup$")
     public void verifySuccesfulOperationsDeploy(String webappName) throws Throwable {
-        jwalaUi.isElementExists(By.xpath("//*[text()='" + webappName + " resource files deployed successfully'"));
```
jeddcuison, line 86: This is to verify that the correct popup is shown before clicking okay. Why remove it ?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/SeleniumTestHelper.java
```diff
@@ -27,38 +41,45 @@
 
     /**
      * Create an instance of a {@link WebDriver} to facilitate browser based testing
+     *
      * @param webDriverClass The name of the web driver class to use
      * @return {@link WebDriver}
      */
-    public static WebDriver createWebDriver(final String webDriverClass) {
+    public static WebDriver createWebDriver(final String webDriverClass) throws IOException {
```
jeddcuison: Please wrap IOException

    throws new SeleniumTestCaseException(...)

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/operation/DrainTest.java
```diff
@@ -1,14 +0,0 @@
-package com.cerner.jwala.ui.selenium.operation;
```
jeddcuison, line 1: Just to confirm, you removed this because you optimized the code to do the drain else where right ? If yes, nice one.

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/operation/IndividualGenerateTest.java
```diff
@@ -1,14 +0,0 @@
-package com.cerner.jwala.ui.selenium.operation;
```
jeddcuison, line 1: Do we need to remove this ?

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/configuration/resourceErrorHandling.feature
```diff
@@ -1,23 +1,23 @@
 Feature: Testing errors caused by invalid template, meta data or file type
 
-  Scenario: Deploy a war file compressed using 7z
```
jeddcuison, line 3: Why remove ? I think this scenario was written to test a bug that we had with regards to incorrect zip.

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/configuration/uploadResource.feature
```diff
@@ -1,75 +1,75 @@
 Feature: Upload a Resource
 
-Scenario: Upload a JVM Resource
+  Scenario: Upload a JVM Resource
 
     Given I logged in
     And I am in the Configuration tab
 
-    # create media
+      # create media
```
jeddcuison: Please align the comment with the code so it is consistent with the rest of the features

    # create  media
    And blah blah blah....

jeddcuison (commented):
I added some comments please check. In general, IE tests are working good job! Some tests were removed, is it because they were redundant and transferred somewhere else ? Some element checking were removed also specifically from the popup,is that necessary ?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/operation/DrainTest.java
Sharvaribarve: yeah drain is included in respective control operations

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/operation/IndividualGenerateTest.java
Sharvaribarve: this is also included individual control operations

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/configuration/resourceErrorHandling.feature
Sharvaribarve: Hi I have included this in another scenario for testing of error in webapp resource, so that we don't duplicate creating jvm and webapp

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/configuration/ManageWsResourceTest.java
Sharvaribarve: Just to reply to your previous comment....external property test is present in the report.
As for this test, I think we have a upload resource file which takes care of all scenarios.

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/operation/DeleteInOperationsTest.java
Sharvaribarve: This is already included in control operations

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/operation/StatusTest.java
Sharvaribarve: I removed it because there is no corresponding feature file for it even in the master however I will add it in control feature

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HandleResourceRunSteps.java
Sharvaribarve: it was weirdly throwing a remote webdriver exception saying that the webdriver has died whenever it had to find this element

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HotDeployRunSteps.java
Sharvaribarve: same case as above

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HotDeployRunSteps.java
Sharvaribarve: same case as above

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/configuration/uploadResource.feature
Sharvaribarve: doing it

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/operation/StatusTest.java
Sharvaribarve: it is disabled in control operations for the following reason
 disabled to make this feature run successfully in Chrome
Chrome replaces http to https but not the non-default (port 80) http port number

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/configuration/uploadResource.feature
Sharvaribarve: done

Sharvaribarve:
The ie branch has the same tests as the master branch except the following
1)popup message checking 3 methods which throws remote webdriver exception 
2)sorting by hostname....this is mainly done because the service removal for linux requires the hostname to be valid and thus we cannot provide arbitrary hostname for sorting

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HandleResourceRunSteps.java
jeddcuison: This needs to be addressed properl. If you are having a hard time looking for a solution you should bring it to the team's attention but don't remove the code just to make it work. It was there to test if the popup is showing the correct message.

Properly addressing the problem may mean that if we can't find a fix we could opt to comment it out instead and put a TODO if the team agrees to it.

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HotDeployRunSteps.java
jeddcuison: same as above...needs to be addressed properly

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HotDeployRunSteps.java
jeddcuison: same as above...needs to be addressed properly

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/configuration/resourceErrorHandling.feature
jeddcuison: Is the war in the scenario you are saying is compressed using 7z also ?

jeddcuison:
"sorting by hostname....this is mainly done because the service removal for linux requires the hostname to be valid and thus we cannot provide arbitrary hostname for sorting"

does this mean that our Chrome test does not work for Linux on sorting ? IF so...

It seems that removing sorting by host name is a fix for Linux for both Chrome and IE. My suggestion for next time is that changes that deviate from the main objective might be better off done in another branch so that the review will be pretty straight forward or if it's just too cumbersome to put it in another branch, just do a separate commit for it and put the appropriate comment e.g. remove blah blah to make blah blah work in Linux.

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/SeleniumTestHelper.java
jeddcuison: Please address my comment above...

jeddcuison (commented):
One of my comments regarding IOException might have been missed. We should discuss what to do with the popup messages validation and not just delete it.

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/configuration/resourceErrorHandling.feature
Sharvaribarve: yeah lines 28-33 and 43-46

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HandleResourceRunSteps.java
Sharvaribarve: Done

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HotDeployRunSteps.java
Sharvaribarve: Done

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/HandleResourceRunSteps.java
```diff
@@ -144,6 +144,13 @@ public void selectResourceCheckBox(String resourceFile) {
 
     @Then("^I confirm delete a resource popup$")
     public void deleteResource() {
+        jwalaUi.isElementExists(By.xpath("//*[contains(text(), 'Are you sure you want to delete the selected resource template(s) ?')]"));
```
jeddcuison: What was the fix on the "weird error" ? Is it using "//*..." instead of "//div..." ? The thing is I noticed that the code you added in HotDeployRunSteps.java at line 80 uses "//div". I guess it works since you used it. I avoid using asterisk since I want my query to be more concise.

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/SeleniumTestHelper.java
```diff
@@ -28,36 +41,45 @@
...
-            throw new SeleniumTestCaseException(e);
-        }
+
+        final DesiredCapabilities dc = DesiredCapabilities.internetExplorer();
+        dc.setCapability(CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR, UnexpectedAlertBehaviour.DISMISS);
+        dc.setCapability(CapabilityType.TAKES_SCREENSHOT, true);
+        dc.setCapability(CapabilityType.SUPPORTS_JAVASCRIPT, true);
+        dc.setCapability(InternetExplorerDriver.INTRODUCE_FLAKINESS_BY_IGNORING_SECURITY_DOMAINS, true);
+        String hubUrl = (String) getProperties().getProperty("hub.url");
+        hubUrl = hubUrl.replaceAll("\"", "");
```
jeddcuison: Missed reviewing this one my bad! This is for selenium Grid I suppose ? So can I make the test run locally if I want to ? Guess I need to find out and change this if necessary. By the way looking at the history I only see my name as the committer. I guess it's because I fixed the merge...anyways this is not mine - the hub.url.

jeddcuison: Looks like Selenium Grid is only used in IE...if I were to code it it would be for both my bad (again) I missed it - will change this

PR #178, opened 2017-11-11
re order postman tests

rahulsayini:
change the order of stop in postman test collection


akinny (approved):
Lets revisit this, I doubt api call order is an issue, in this case

jmahmoodcc (approved):

PR #181, opened 2017-11-15
Postman add wait for asynchronous calls

rahulsayini:
Add wait script between asynchronous calls for stop and start of webserver, jvm and group

### DIFF
--- a/jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
+++ b/jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
@@ -2731,7 +2731,17 @@
 							"script": {
 								"type": "text/javascript",
 								"exec": [
-									"tests[\"Status code is 200\"] = responseCode.code === 200;"
+									"tests[\"Status code is 200\"] = responseCode.code === 200;",
+									"",
+									"function pausecomp(millis)",
+									" {",
+									"  var date = new Date();",
+									"  var curDate = null;",
+									"  do { curDate = new Date(); }",
+									"  while(curDate-date < millis);",
+									"}",
+									"",
+									"pausecomp(10000);"
 								]
 							}
 						}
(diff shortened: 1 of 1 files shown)


jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -2731,7 +2731,17 @@
...
 								"exec": [
-									"tests[\"Status code is 200\"] = responseCode.code === 200;"
+									"tests[\"Status code is 200\"] = responseCode.code === 200;",
+									"",
+									"function pausecomp(millis)",
+									" {",
+									"  var date = new Date();",
+									"  var curDate = null;",
+									"  do { curDate = new Date(); }",
+									"  while(curDate-date < millis);",
```
akinny: This is tight loop, I think it will cause 100% cpu utilization, did you check?

akinny (changes requested):

jmahmoodcc (changes requested):
After the timeout expires you need to query the state of the JVM/Web Server to verify that it is started or stopped like the test intended

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -2733,15 +2733,12 @@
...
 									"tests[\"Status code is 200\"] = responseCode.code === 200;",
 									"",
-									"function pausecomp(millis)",
+									"function mockSleep(millis)",
 									" {",
-									"  var date = new Date();",
-									"  var curDate = null;",
-									"  do { curDate = new Date(); }",
-									"  while(curDate-date < millis);",
+									"return new Promise(resolve => setTimeout(resolve, millis))",
```
jmahmoodcc: Promises and setTimeout are non-blocking - they execute asynchronously but they don't pause the thread. This solution won't have the intended delay that you're looking for.

See the answer here for more explanation:
https://stackoverflow.com/questions/36588775/are-javascript-promise-asynchronous

jmahmoodcc (changes requested):
Using the promise and setTimeout is non-blocking, they won't actually pause the thread. See my comment for more details.

akinny:
Removing tight loop

PR #183, opened 2017-11-16
WIP: Add client name to jwala home page

rahulsayini:
Request to have Client name appear in the jwala UI


Sharvaribarve (changes requested):
I think the jira requests for more data than just the name,
and the format wanted is a banner.

LinHungWu (approved):

PR #188, opened 2017-12-12
Client name hhrr datamode

rahulsayini:
Add client HHRR and Data mode to jwala UI

### DIFF
--- a/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/admin/AdminRunSteps.java
+++ b/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/admin/AdminRunSteps.java
@@ -51,10 +51,8 @@ public class AdminRunSteps {
         assert StringUtils.isNotEmpty(val) && val.contains(itemName);
     }
 
-    @And("^I see client details \"(.*)\" \"(.*)\" \"(.*)\"")
-    public void verifyClientName(final String clientName, final String hHRR, final String dataMode){
+    @And("^I see client details \"(.*)\"")
+    public void verifyClientName(final String clientName){
         jwalaUi.find(By.xpath("//span[text() = '" + paramProp.getProperty(clientName) + "']"));
-        jwalaUi.find(By.xpath("//span[text() = '" + paramProp.getProperty(hHRR) + "']"));
-        jwalaUi.find(By.xpath("//span[text() = '" + paramProp.getProperty(dataMode) + "']"));
     }
 }
--- a/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/admin/AdminRunSteps.java
+++ b/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/admin/AdminRunSteps.java
@@ -51,8 +51,9 @@ public class AdminRunSteps {
         assert StringUtils.isNotEmpty(val) && val.contains(itemName);
     }
 
-    @And("^I see client details \"(.*)\"")
-    public void verifyClientName(final String clientName){
-        jwalaUi.find(By.xpath("//span[text() = '" + paramProp.getProperty(clientName) + "']"));
+    @And("^I see client details \"(.*)\" \"(.*)\"")
+    public void verifyClientDetails(final String clientDetails, final String dataMode){
+        jwalaUi.find(By.xpath("//span[text() = '" + paramProp.getProperty(clientDetails) + "']"));
+        jwalaUi.find(By.xpath("//span[text() = '" + paramProp.getProperty(dataMode) + "']"));
     }
 }
--- a/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/admin/AdminRunSteps.java
+++ b/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/admin/AdminRunSteps.java
@@ -50,4 +50,11 @@ public class AdminRunSteps {
         final String val = webElement.getAttribute("value");
         assert StringUtils.isNotEmpty(val) && val.contains(itemName);
     }
+
+    @And("^I see client Name \"(.*)\" \"(.*)\" \"(.*)\"")
+    public void verifyClientName(final String clientName, final String hHRR, final String dataMode){
+        jwalaUi.find(By.xpath("//span[text() = '" + paramProp.getProperty(clientName) + "']"));
+        jwalaUi.find(By.xpath("//span[text() = '" + paramProp.getProperty(hHRR) + "']"));
+        jwalaUi.find(By.xpath("//span[text() = '" + paramProp.getProperty(dataMode) + "']"));
+    }
 }
--- a/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/admin/AdminRunSteps.java
+++ b/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/admin/AdminRunSteps.java
@@ -51,7 +51,7 @@ public class AdminRunSteps {
         assert StringUtils.isNotEmpty(val) && val.contains(itemName);
     }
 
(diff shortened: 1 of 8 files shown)


jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/admin/admin.feature
```diff
@@ -3,6 +3,7 @@ Feature: Admin Utilities
 Scenario: Test encryption, properties reload and manifest.mf
 
     Given I logged in
+    And I see client Name "jwala.client.name" "jwala.HHRR" "jwala.data.mode"
```
Sharvaribarve: Could you make this line more verbose?

jwala-webapp/src/main/java/com/cerner/jwala/web/javascript/variable/property/ApplicationPropertySourceDefinition.java
```diff
@@ -18,17 +18,20 @@
     JWALA_ROLE_ADMIN("jwalaRoleAdmin", "jwala.role.admin", "Tomcat Admin", VariableStyle.STRING),
     TOMCAT_MANAGER_PROTOCOL("tomcatManagerProtocol", "tomcat.manager.protocol", "http", VariableStyle.STRING),
     APACHE_HTTPD_STATUS_PING_RESOURCE("apacheHttpdStatusPingResource", "apache.httpd.status.ping.resource", "index.html", VariableStyle.STRING),
-    TOMCAT_IMAGE_LOGO_PATH("tomcatImageLogoPath", "tomcat.image.logo.path", "/tomcat-power.gif", VariableStyle.STRING);
+    TOMCAT_IMAGE_LOGO_PATH("tomcatImageLogoPath", "tomcat.image.logo.path", "/tomcat-power.gif", VariableStyle.STRING),
+    JWALA_CLIENT_NAME("jwalaClientName", "jwala.client.name", "Cerner Corporation", VariableStyle.STRING),
+    JWALA_HHRR("jwalaHhrr","jwala.hhrr", "B007", VariableStyle.STRING),
```
Sharvaribarve: Could the default value be blank instead of B007 😆

jeddcuison: +1

jeddcuison: Actually after thinking about @Sharvaribarve's comment, do we need default values for client name as well ? Maybe we can ask the team.

jwala-webapp/src/main/webapp/resources/css/react/main-area.css
```diff
@@ -10,6 +10,13 @@ div.main-area {
     margin-right:5px;
 }
 
+.association-name {
+    position: relative;
+    top: 80px;
+	float:right;
```
jeddcuison: I know this might be knit picking but this should be aligned.

jwala-webapp/src/main/java/com/cerner/jwala/web/javascript/variable/property/ApplicationPropertySourceDefinition.java
rahulsayini: Yes, I agree to making all fields default to blank. I will ask team before doing it.

jwala-tomcat/src/main/resources/data/properties/vars.properties
```diff
@@ -56,4 +56,9 @@ jwala.binary.dir=../data/binaries
 jwala.agent.dir=../data/agent
 
 #jsch verbose logging flag
-ssh.verbose=false
\ No newline at end of file
+ssh.verbose=false
+
+#jwala client name
```
jmahmoodcc, line 61: To me, it appears that these parameters should be optional and we shouldn't provide default values here.
At the least, I would provide the keys with empty values and then have them commented out with a comment providing a description of where those values would get displayed when provided.

jmahmoodcc (changes requested):
The HHRR and Data Mode fields are very specific to Cerner solutions. For the more generic solution we should really only provide a single key that contains the entire name. Then in our chef cookbook, where we provide our own vars.properties template, we should construct that string from the HHRR and Data Mode provided in the environment json.

jwala-webapp/src/test/java/com/cerner/jwala/web/javascript/variable/property/ApplicationPropertySourceDefinitionTest.java
```diff
@@ -20,4 +20,11 @@ public void testLoadBalancerStatusMount() {
...
     }
 
+    @Test
+    public void testJwalaClientDetails() {
+        assertEquals("", JWALA_CLIENT_NAME.getDefaultValue());
+        assertEquals("", JWALA_HHRR.getDefaultValue());
+        assertEquals("", JWALA_DATA_MODE.getDefaultValue());
+    }
+
 }
```
akinny: Add one more unit test with values

jwala-integration-test/src/test/resources/selenium/test.properties
```diff
@@ -32,4 +32,5 @@ httpd.resource.deploy.path=d:/ctp/app/data/httpd
 webapp.resource.deploy.path=d:/ctp/app/webapps
 webapp.context.resource.deploy.path=${jvm.tomcatMedia.remoteDir}/${jvm.jvmName}/${jvm.tomcatMedia.rootDir}/conf/stp/localhost
 jvm.setenv.resource.deploy.path=${jvm.tomcatMedia.remoteDir}/${jvm.jvmName}/${jvm.tomcatMedia.rootDir}/bin
-jvm.server.xml.resource.deploy.path=${jvm.tomcatMedia.remoteDir}/${jvm.jvmName}/${jvm.tomcatMedia.rootDir}/conf
\ No newline at end of file
+jvm.server.xml.resource.deploy.path=${jvm.tomcatMedia.remoteDir}/${jvm.jvmName}/${jvm.tomcatMedia.rootDir}/conf
+jwala.client.name=Cerner Corporation B007 DEV
```
Sharvaribarve: If you are testing only the name, why have the other field values?

rahulsayini: As per other code review comments we are only using one field which shows client name. But, our solution requirement needs HHRR and datamode So, i am passing them from chef template. Now, i need to test all three values for which i can only pass from test.properties.

akinny: Sorry, I missed this caught my eye when I was looking at the chef code. Add 3 properties then concatenate them in Jwala java code

akinny (changes requested):
I think we should have 3 properties

jmahmoodcc (approved):

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/admin/AdminRunSteps.java
```diff
@@ -50,4 +50,10 @@ public void verifyTextAreaItem(final String itemName, final String textAreaHeadi
         final String val = webElement.getAttribute("value");
         assert StringUtils.isNotEmpty(val) && val.contains(itemName);
     }
+
+    @And("^I see client details \"(.*)\" \"(.*)\"")
```
jeddcuison: Just make the statement more precise e.g. 

    I see in the web banner that the client name = \"(.*)\", HHRR = \"(.*)\" and data mode = \"(.*)\"

With regards to the HHRR let's put a sample HHRR for our 49 environment

jeddcuison (commented):
Please see my comments thanks

jwala-tomcat/src/main/resources/data/properties/vars.properties
jeddcuison: done by @rahulsayini

jeddcuison (approved):

PR #192, opened 2017-12-14
Postman add wait for asynchronous calls

rahulsayini:
Add set time out between start and stop API calls


jmahmoodcc (approved):

jmahmoodcc:
I'm so glad setTimeout() worked - although I still don't fully understand why.
Sorry folks if I steered you in the wrong direction. That's my fault for making this overly complicated.

akinny (approved):

PR #195, opened 2018-01-23
Thread dump logging

Sharvaribarve:


jwala-common/src/main/java/com/cerner/jwala/common/request/jvm/ThreadDumpControlJvmRequest.java
```diff
@@ -0,0 +1,22 @@
+package com.cerner.jwala.common.request.jvm;
+
+import com.cerner.jwala.common.domain.model.jvm.Jvm;
+import com.cerner.jwala.common.domain.model.jvm.JvmControlOperation;
+
+/**
+ *
```
akinny: Can you add some documentation at class & method level

jwala-services/src/main/java/com/cerner/jwala/control/jvm/command/JvmCommandFactory.java
```diff
@@ -198,10 +198,13 @@ private ExecCommand getExecCommandForHeapDump(String scriptName, Jvm jvm) {
      * @return
      */
     private ExecCommand getExecCommandForThreadDump(String scriptName, Jvm jvm) {
+        final String trimmedJvmName = StringUtils.deleteWhitespace(jvm.getJvmName());
         final String jvmRootDir = Paths.get(jvm.getTomcatMedia().getRemoteDir().toString() + '/' +
                 StringUtils.deleteWhitespace(jvm.getJvmName()) + '/' + jvm.getTomcatMedia().getRootDir())
                 .normalize().toString();
-        return new ExecCommand(getFullPathScript(jvm, scriptName), jvm.getJavaHome(), jvmRootDir, jvm.getJvmName());
+        final DateTimeFormatter formatter = DateTimeFormat.forPattern("yyyyMMdd.HHmmss");
+        final String dumpFile = "threadDump." + trimmedJvmName + "." + formatter.print(DateTime.now());
```
akinny: I missed this before. Can you rename the generated file and add an extension .txt to it, since most of our users are windows users?
Rename
thread_dump_01-24-2018-12.21.23
to something like
thread-dump-01-24-2018-12-21-23.txt

akinny (changes requested):
Minor changes, overall looks good

jwala-tomcat/src/main/resources/data/scripts/thread-dump.sh
```diff
@@ -8,6 +8,7 @@ esac
 export JAVA_HOME=$1
 export JVM_INSTANCE_DIR=$2
 export JVM_NAME=$3
+export DUMP_FILE=$4
 
 if [ ! -d ${JVM_INSTANCE_DIR}/thread_dump_reports} ]; then
       mkdir -p ${JVM_INSTANCE_DIR}/thread_dump_reports
```
akinny: Make this directory consistent with heap dumps

jwala-tomcat/src/main/resources/data/scripts/thread-dump.sh
```diff
@@ -8,19 +8,28 @@ esac
...
 current_time=$(date +"%m-%d-%Y-%T")
 
 if $linux; then
 	echo $(<${JVM_INSTANCE_DIR}/logs/catalina.pid)
-	/usr/bin/sudo -u tomcat ${JAVA_HOME}/bin/jstack $(<${JVM_INSTANCE_DIR}/logs/catalina.pid) 2>&1 | tee  ${JVM_INSTANCE_DIR}/thread_dump_reports/thread_dump_$current_time
+	thread_dump_output=$(/usr/bin/sudo -u tomcat ${JAVA_HOME}/bin/jstack $(<${JVM_INSTANCE_DIR}/logs/catalina.pid) 2>&1 )
+    if echo "$thread_dump_output" | grep -iqF "Full thread dump"; then
+	   echo "$thread_dump_output" |tee ${DATA_DIR}/$DUMP_FILE
+	   echo "Creating thread-dump file at location: ${DATA_DIR}/$DUMP_FILE, "
+	fi
```
jmahmoodcc: Add an else clause here with a message stating that the thread dump failed or didn't execute as expected. You might want to include the content of thread_dump_output as well for debugging purposes.

jwala-tomcat/src/main/resources/data/scripts/thread-dump.sh
```diff
@@ -8,19 +8,28 @@ esac
...
 if $cygwin; then
     export JVMPID=`sc queryex $JVM_NAME | /usr/bin/grep PID | /usr/bin/awk '{ print $3 }'`
-	${JAVA_HOME}/bin/jstack -l ${JVMPID} 2>&1 |tee ${JVM_INSTANCE_DIR}/thread_dump_reports/thread_dump_$current_time
-fi
\ No newline at end of file
+    thread_dump_output=$(${JAVA_HOME}/bin/jstack -l ${JVMPID} 2>&1 )
+    if echo "$thread_dump_output" | grep -iqF "Full thread dump"; then
+       echo "$thread_dump_output" |tee ${DATA_DIR}/$DUMP_FILE
+	   echo "Creating Thread Dump file at location : ${DATA_DIR}/$DUMP_FILE, "
+	fi
```
jmahmoodcc: Same thing here. If the success condition isn't met, then by default it failed - but we won't know why because we don't capture any of the unexpected content.

jmahmoodcc (changes requested):
Looks good, just some minor comments in the thread_dump.sh

jmahmoodcc (changes requested):
I just noticed one more thing after you added the error messages. Each clause should return an appropriate error condition: exit 0 for success and 1 for failure. 

The return code should be checked by the java service when the ssh command completes.

akinny (approved):

jwala-tomcat/src/main/resources/data/scripts/thread-dump.sh
```diff
@@ -8,19 +8,38 @@ esac
...
     export JVMPID=`sc queryex $JVM_NAME | /usr/bin/grep PID | /usr/bin/awk '{ print $3 }'`
-	${JAVA_HOME}/bin/jstack -l ${JVMPID} 2>&1 |tee ${JVM_INSTANCE_DIR}/thread_dump_reports/thread_dump_$current_time
-fi
\ No newline at end of file
+    thread_dump_output=$(${JAVA_HOME}/bin/jstack -l ${JVMPID} 2>&1 )
+    THREAD_DUMP_RE  TURN_CODE=$?
+    if echo "$thread_dump_output" | grep -iqF "Full thread dump"; then
+       echo "$thread_dump_output" |tee ${DATA_DIR}/$DUMP_FILE
+	   echo "Creating Thread Dump file at location : ${DATA_DIR}/$DUMP_FILE, "
+	   exit $JWALA_EXIT_CODE_SUCCESS
```
jmahmoodcc: Why use 2 different ways of returning the status code?
For the linux thread dump the exit status is stored in the THREAD_DUMP_RETURN_CODE, but here for the cygwin way the status code is returned as a constant. Since you check specifically for the "Full thread dump" substring I would use the explicit return of the constant for both ways - that is, change the linux way to return the constant and don't even check for the jstack return value.

Also on line 34, the THREAD_DUMP_RETURN_CODE has a space in the constant name and is never used so get rid of it anyway.

jmahmoodcc (changes requested):
Close, but still needs a little clean up.

jwala-services/src/main/java/com/cerner/jwala/control/jvm/command/JvmCommandFactory.java
```diff
@@ -203,7 +204,8 @@ private ExecCommand getExecCommandForThreadDump(String scriptName, Jvm jvm) {
                 StringUtils.deleteWhitespace(jvm.getJvmName()) + '/' + jvm.getTomcatMedia().getRootDir())
                 .normalize().toString();
         final DateTimeFormatter formatter = DateTimeFormat.forPattern("yyyyMMdd.HHmmss");
-        final String threadDumpDir = jvm.getTomcatMedia().getRemoteDir().normalize().toString() + "/" + jvm.getJvmName();
+        final String threadDumpDir = Paths.get(jvm.getTomcatMedia().getRemoteDir().toString() + "/../" + "data")
```
jmahmoodcc: This path assumes that Jwala is running in a Cerner installed environment, which will most likely not be the case for the rest of the open source community.
Since this path could be anything specified by the user, backing up to a parent directory that may not exist would throw an exception when attempting any kind of operation.

Why not keep it in the JVM directory like before?

jmahmoodcc (changes requested):
See my comment about the thread dump path

jwala-services/src/main/java/com/cerner/jwala/control/jvm/command/JvmCommandFactory.java
Sharvaribarve: The jira requests the path of the file to be in data directory

jmahmoodcc (approved):

PR #196, opened 2018-01-31
Fix linux selenium

Sharvaribarve:
Added support for linux and exception handling for invalid hostname


akinny (approved):

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/TearDownStep.java
```diff
@@ -83,15 +98,50 @@ private void deleteService(String sshUser, String sshPwd, ServiceInfo serviceInf
         final RemoteSystemConnection remoteSystemConnection
                 = new RemoteSystemConnection(sshUser, sshPwd, serviceInfo.host, 22);
 
+        RemoteCommandReturnInfo remoteCommandReturnInfo = null;
+        try {
+            remoteCommandReturnInfo = jschService.runShellCommand(remoteSystemConnection, "uname", 10000);
```
jmahmoodcc: Make 10000 a constant or a property

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/TearDownStep.java
```diff
@@ -83,15 +98,50 @@ private void deleteService(String sshUser, String sshPwd, ServiceInfo serviceInf
...
+        final JwalaOsType osType = StringUtils.indexOf(remoteCommandReturnInfo.standardOuput, "CYGWIN") > -1 ?
+                JwalaOsType.WINDOWS : JwalaOsType.UNIX;
+        LOGGER.info("OS type = {}", osType);
+
         if (serviceInfo.isStarted()) {
             // Stop the service first so that service delete is executed by the OS asap
-            jschService.runShellCommand(remoteSystemConnection, "sc stop " + serviceInfo.name, 300000);
+            switch (osType) {
+                case WINDOWS:
+                    remoteCommandReturnInfo = jschService.runShellCommand(remoteSystemConnection, "sc stop " + serviceInfo.name, 300000);
```
jmahmoodcc: Make 30000 a constant or a property

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/TearDownStep.java
```diff
@@ -83,15 +98,50 @@ private void deleteService(String sshUser, String sshPwd, ServiceInfo serviceInf
...
+                    break;
+                case UNIX:
+                    remoteCommandReturnInfo = jschService.runShellCommand(remoteSystemConnection, "sudo service " + serviceInfo.name + " stop", 300000);
```
jmahmoodcc: Again, make "magic numbers" constants or properties

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/TearDownStep.java
```diff
@@ -83,15 +98,50 @@ private void deleteService(String sshUser, String sshPwd, ServiceInfo serviceInf
...
+                            MessageFormat.format("I don't know how to stop services for os type -> {0}", osType));
+            }
+            LOGGER.info("Shell command executed = {}", remoteCommandReturnInfo.standardOuput);
         }
 
         // If the service state is NEW, that means that the service has not yet been created so let's not issue a
         // delete command to save some time
         if (!serviceInfo.isNew()) {
-            jschService.runShellCommand(remoteSystemConnection, "sc delete " + serviceInfo.name, 300000);
+            switch (osType) {
```
jmahmoodcc, line 144: Even though the stop command was issued above, we should verify that the JVM is indeed stopped before deleting the service

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/TearDownStep.java
```diff
@@ -83,15 +98,50 @@ private void deleteService(String sshUser, String sshPwd, ServiceInfo serviceInf
...
+                case WINDOWS:
+                    remoteCommandReturnInfo = jschService.runShellCommand(remoteSystemConnection, "sc delete " + serviceInfo.name, 300000);
+                    break;
+                case UNIX:
+                    remoteCommandReturnInfo = jschService.runShellCommand(remoteSystemConnection, "sudo rm /etc/init.d/" + serviceInfo.name, 300000);
```
jmahmoodcc: Deleting the Linux service also requires calling 'chkconfig --del' to clean up any run level definitions for the service

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/TearDownStep.java
```diff
@@ -83,15 +98,50 @@ private void deleteService(String sshUser, String sshPwd, ServiceInfo serviceInf
...
+            LOGGER.info("Shell command executed = {}", remoteCommandReturnInfo.standardOuput);
         }
 
         // If the service state is NEW, that means that the service has not yet been created so let's not issue a
         // delete command to save some time
         if (!serviceInfo.isNew()) {
-            jschService.runShellCommand(remoteSystemConnection, "sc delete " + serviceInfo.name, 300000);
+            switch (osType) {
+                case WINDOWS:
+                    remoteCommandReturnInfo = jschService.runShellCommand(remoteSystemConnection, "sc delete " + serviceInfo.name, 300000);
```
jmahmoodcc: Is there any checking of the status returned from the remote command in case of errors?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/admin/AdminRunSteps.java
```diff
@@ -53,7 +53,9 @@ public void verifyTextAreaItem(final String itemName, final String textAreaHeadi
 
     @And("I see in the web banner that the client details = \"(.*)\" and data mode = \"(.*)\"")
     public void verifyClientDetails(final String clientDetails, final String dataMode){
-        jwalaUi.find(By.xpath("//span[text() = '" + paramProp.getProperty(clientDetails) + "']"));
-        jwalaUi.find(By.xpath("//span[text() = '" + paramProp.getProperty(dataMode) + "']"));
+        if(paramProp.getProperty("os").equals("windows")) {
```
jmahmoodcc: Why is this only checked in windows? It should work the same for both Linux and Windows

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/CreateJvmRunSteps.java
```diff
@@ -86,7 +93,7 @@ public void checkForJvm(final String jvmName) {
     @Then("^I see the following jvm details in the jvm table:$")
     public void validateJvm(final Map<String, String> jvmDetails) {
         jwalaUi.waitUntilElementIsVisible(By.xpath("//button[text()='" + jvmDetails.get("name") + "']"));
-        assertTrue(jwalaUi.isElementExists(By.xpath("//td[text()='" + jvmDetails.get("host") + "']")));
+        assertTrue(jwalaUi.isElementExists(By.xpath("//td[text()='" + getPropertyValue(jvmDetails.get("host")) + "']")));
```
jmahmoodcc, line 96: This looks like it's reading a property where the key is the JVM's host. That doesn't seem right ... I would expect the key to be something like 'host' or 'hostname'

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/CreateJvmRunSteps.java
```diff
@@ -99,4 +106,8 @@ public void validateJvm(final Map<String, String> jvmDetails) {
     public void waitForJvm(String name) {
         jwalaUi.waitUntilElementIsVisible(By.xpath("//button[text()='" + name + "']"));
     }
+
+    private String getPropertyValue(final String key) {
+        return paramProp.getProperty(key) == null ? key : paramProp.getProperty(key);
```
jmahmoodcc, line 111: Why return the key if the property is not found? 
It would be better to return null or an empty string.
Or, if the key is required to be set as a property then I would actually throw an exception.

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/configuration/manageWebServer.feature
```diff
@@ -14,7 +14,7 @@ Scenario: Add Web Server
     When I click the add web server button
     And I see the web server add dialog
     And I fill in the "Web Server Name" field with "WEBSERVER_X"
-    And I fill in the "Host Name" field with "localhost"
+    And I fill in the "Host Name" field with "host1"
```
jmahmoodcc, line 17: In looking at this code now, I wonder: Is there a way to make these tests in the feature more readable? It feels like it defeats the purpose to have the value in the test point to a key in a property file some where. 
Is it possible to dynamically generate the feature from the property file, and then run the test from that generated feature. Then the test would be more readable, and the code base wouldn't require all the checks when reading a property.

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/operations/controlJvm.feature
```diff
@@ -1,79 +1,79 @@
...
 
     When I click the drain button of "CONTROL-JVM-TEST-J" JVM under "CONTROL-JVM-TEST-G" group
     And I do not see an error message after clicking drain
 
-    # thread dump
+      # thread dump
     When I click thread dump of jvm "CONTROL-JVM-TEST-J" of the group "CONTROL-JVM-TEST-G"
     Then I don't see the click status tooltip
-    And I see the thread dump popup
+    And I see the thread dump popup for the jvm "CONTROL-JVM-TEST-J" with the message "thread-dump-popup-message"
```
jmahmoodcc: This is another example here - I feel like this test should be written with the expected message in the pop up and not some key to a property

jmahmoodcc (changes requested):
Some minor changes, but also some major questions - not necessarily issues, but ... questions ...

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/admin/AdminRunSteps.java
Sharvaribarve: This feature of client name  is only available in windows currently...

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/configuration/CreateJvmRunSteps.java
Sharvaribarve: actually no....imagine in case of sorting I have a hostname as aaaHost, in this case I would use this itself as the properhost name since there is no value present in the properties file

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/configuration/manageWebServer.feature
Sharvaribarve: unfortunately in most of the cases...like this we need to have correct hostnames, and for security reasons we cannot have those in open source

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/TearDownStep.java
Sharvaribarve: ok will add that

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/TearDownStep.java
Sharvaribarve: done

PR #200, opened 2018-02-12
Delete linux service

jmahmoodcc:
The fix requires 2 changes:
- remove the 'test' keyword when checking the existing of the init.d service script
- add a call to clean up the run level setting for the service (chkconfig)

Also added the line to clean up the run level setting in the Selenium tear down step.

### DIFF
--- a/jwala-tomcat/src/main/resources/data/scripts/delete-service.sh
+++ b/jwala-tomcat/src/main/resources/data/scripts/delete-service.sh
@@ -22,7 +22,9 @@ fi
 
 if $linux; then
     if $linux; then
-        if [ test -e "/etc/init.d/$1" ]; then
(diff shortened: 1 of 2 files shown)


jwala-tomcat/src/main/resources/data/scripts/delete-service.sh
```diff
@@ -22,9 +22,11 @@ fi
 
 if $linux; then
```
akinny, line 23: Can you remove one if check?

jmahmoodcc: Sure, I guess we just wanted to make sure it was really Linux :)

akinny (approved):

Sharvaribarve (approved):

PR #202, opened 2018-02-16
Address some tech debt

jeddcuison:
These commits addresses the 6 critical issues listed in Sonar

### DIFF
--- a/jwala-common/src/main/java/com/cerner/jwala/common/FileUtility.java
+++ b/jwala-common/src/main/java/com/cerner/jwala/common/FileUtility.java
@@ -216,12 +216,12 @@ public class FileUtility {
             return DigestUtils.sha256Hex(fileInputStream);
         } catch (FileNotFoundException e) {
             final String errMsg = MessageFormat.format("Failed to get the checksum for non-existent file: {0}", fileAbsolutePath);
-            LOGGER.error(errMsg);
-            throw new FileUtilityException(errMsg);
+            LOGGER.error(errMsg, e);
+            throw new FileUtilityException(errMsg, e);
         } catch (IOException e) {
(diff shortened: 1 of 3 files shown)


jwala-common/src/main/java/com/cerner/jwala/common/FileUtility.java
```diff
@@ -100,7 +99,7 @@ private void closeJarFile(final JarFile jarFile) {
     public Set<String> getZipRootDirs(final String zipFilename) {
         final Set<String> zipRootDirs = new HashSet<>();
         try {
-            if (zipFilename.indexOf(TAR_GZIP_FILE_EXT) > 0) {
+            if (TAR_GZIP_FILE_EXT.equalsIgnoreCase(FilenameUtils.getExtension(zipFilename))) {
```
jmahmoodcc, line 102: This logic worries me because it's not a 1 for 1 replacement. The original check was looking for any instance of ".gz" which in itself seems error prone, but the new logic checks for a file extension of "gz".
Did you regression test this change already on a Linux deployment?

Or even better, is there a way to check the file format with the Tika library?

jmahmoodcc (commented):
Just one question about how we check for compressed files with a gz format

jwala-common/src/main/java/com/cerner/jwala/common/FileUtility.java
jeddcuison: It's not a 1 to 1 replacement because FilenameUtils.getExtension returns the extension without the dot. e.g. test.tar.gz will return "gz" but test.targz will return blank.

jeddcuison: Good suggestion on TIKA, it does support detection of gz file. I'll change the code to use gz and write the tests for it.

PR #203, opened 2018-02-20
Add Selenium tests for checking the JVM and Web Server service was deleted

jmahmoodcc:
These changes are for updating the Selenium tests to check that the Windows and Linux services no longer exist after a JVM or Web Server is deleted.


jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/SeleniumTestHelper.java
```diff
@@ -131,4 +121,78 @@ public static void runSqlScript(final String sqlScript) throws IOException, Clas
             }
         }
     }
+
+    public static void checkServiceDeleteWasSuccessful(String serviceName, JschService jschSvc, Properties properties) {
+        LOGGER.info("checkServiceDeleteWasSuccessful {}", serviceName);
+
+        props = properties;
```
jeddcuison, line 128: First time I saw this it confused me. I think writing it this way makes it easier to read.

    this.props = props;
    this.jschService = jschService;

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/SeleniumTestHelper.java
```diff
@@ -131,4 +121,78 @@ public static void runSqlScript(final String sqlScript) throws IOException, Clas
...
+        System.setProperty("PROPERTIES_ROOT_PATH", SeleniumTestHelper.class.getResource("/selenium/vars.properties").getPath()
+                .replace("/vars.properties", ""));
+
+        final String hostname = props.getProperty("host1");
+
+        final RemoteSystemConnection remoteSystemConnection = getRemoteSystemConnection(hostname);
+
+        final JwalaOsType osType = getJwalaOsType(remoteSystemConnection);
+
+        if (osType.equals(JwalaOsType.WINDOWS)) {
```
jeddcuison, line 142: To be safe from NPE this should be written as

    if (JwalaOsType.WINDOWS.equals(osType)) {
        // do something here...
    }

I know the getJwalaOsType will not return NULL that is for now. We will never know if someone in the future will change the method which would then return null then this code breaks.

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/SeleniumTestHelper.java
```diff
@@ -131,4 +121,78 @@ public static void runSqlScript(final String sqlScript) throws IOException, Clas
...
+            checkWindowsService(serviceName, hostname, remoteSystemConnection);
+        } else {
+            checkLinuxServiceRunLevel(serviceName, hostname, remoteSystemConnection);
+            checkLinuxService(serviceName, hostname, remoteSystemConnection);
+        }
+
+        // Restore the properties root path
+        if (null != originalPropertiesRootPath) {
+            System.setProperty("PROPERTIES_ROOT_PATH", originalPropertiesRootPath);
```
jeddcuison, line 151: I think we should put this in a finally block.

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/SeleniumTestHelper.java
```diff
@@ -131,4 +121,78 @@ public static void runSqlScript(final String sqlScript) throws IOException, Clas
...
+        }
+    }
+
+    private static void checkLinuxService(String serviceName, String hostname, RemoteSystemConnection remoteSystemConnection) {
+        RemoteCommandReturnInfo remoteCommandReturnInfo = jschService.runShellCommand(remoteSystemConnection, "sudo service " + serviceName + " status", SHORT_CONNECTION_TIMEOUT);
```
jeddcuison, line 156: Can the lines be shortened ? Not 80 but enough for it to fit github review screen perhaps ?

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/SeleniumTestHelper.java
```diff
@@ -131,4 +121,78 @@ public static void runSqlScript(final String sqlScript) throws IOException, Clas
...
+        return StringUtils.indexOf(remoteCommandReturnInfo.standardOuput, "CYGWIN") > -1 ? JwalaOsType.WINDOWS : JwalaOsType.UNIX;
+    }
+
+    private static RemoteSystemConnection getRemoteSystemConnection(String hostname) {
+        final String sshUser = props.getProperty("ssh.user.name");
+        final String sshPwd = props.getProperty("ssh.user.pwd");
+
+        LOGGER.info("sshUser {} :: host1: {}", sshUser, hostname);
+
+        return new RemoteSystemConnection(sshUser, sshPwd, hostname, 22);
```
jeddcuison, line 194: 22 should also be a property

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/SeleniumTestHelper.java
jeddcuison: I'm not really in favor of mixing static methods and class variables. If it can't be avoided we might just as well write a service class that does the JVM and web server deletion so that it adheres to the [single responsibility pattern](https://en.wikipedia.org/wiki/Single_responsibility_principle).

jeddcuison (changes requested):
There are some comments that are minor but please check NPE possibility and use of finally block to restore the original properties path.

jmahmoodcc:
@jeddcuison Thanks for the review! I'll make all of the changes as soon as I can

PR #204, opened 2018-02-20
Added retries for httpd.conf

Sharvaribarve:
Added 3 retries for httpd.conf, since otherwise it gives an intermittent error of the browser not rendering the text

### DIFF
--- a/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/operation/WebServerControlRunSteps.java
+++ b/jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/operation/WebServerControlRunSteps.java
@@ -48,7 +48,7 @@ public class WebServerControlRunSteps {
     }
 
     @When("^I click the httpd.conf link of web server \"(.*)\" under group \"(.*)\" in the operations tab and I see httpd.conf$")
-    public void clickWebServerHttpdConfLink(final String linkLabel, final String webServerName, final String groupName) {
+    public void clickWebServerHttpdConfLink(final String webServerName, final String groupName) {
         origWindowHandle = jwalaUi.getWebDriver().getWindowHandle();
         int count = 0;
         while(count < 3) {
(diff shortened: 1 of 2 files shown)


jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/operation/WebServerControlRunSteps.java
```diff
@@ -46,6 +47,26 @@ public void clickWebServerControlLink(final String linkLabel, final String webSe
                 + webServerName + "']/following-sibling::td//button[text()='" + linkLabel + "']"));
     }
 
+    @When("^I click the httpd.conf link of web server \"(.*)\" under group \"(.*)\" in the operations tab and I see httpd.conf$")
+    public void clickWebServerHttpdConfLink(final String webServerName, final String groupName) {
+        origWindowHandle = jwalaUi.getWebDriver().getWindowHandle();
+        int count = 0;
+        while(count < 3) {
```
jmahmoodcc, line 54: We should make the max number of retries a configurable property with a default value of 3

jwala-integration-test/src/test/java/com/cerner/jwala/ui/selenium/steps/operation/WebServerControlRunSteps.java
```diff
@@ -60,13 +81,14 @@ public void verifyIfDeleteWebServerIsSuccessful(final String webServerName) {
     }
 
     @Then("^I see the httpd.conf$")
-    public void verifyProperConfFile() {
+    public boolean verifyProperConfFile() throws TimeoutException{
```
jmahmoodcc, line 84: If this method is no longer used in any other feature we should make it private and remove the @Then annotation

jmahmoodcc (commented):
I'm not sure how I feel about this fix in general - to me it looks like we're masking an intermittent issue. What's the consensus from the rest of the team?

jeddcuison:
Not to fan the flame or anything, I agree with Jeff that this might be an issue and we should probably fix it. Does it happen in Jwala ? Like when you click httpd.conf manually it does not show up sometimes ?

PR #205, opened 2018-02-27
add health check tests for load balancer

rahulsayini:
add tests to check load balancer

### DIFF
--- a/jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
+++ b/jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
@@ -7182,6 +7182,128 @@
 						"description": ""
 					},
 					"response": []
+				}
+			]
+		},
+		{
+			"name": "u) health check application test",
+			"description": "",
+			"item": [
+				{
+					"name": "start jvm with wait true",
+					"event": [
+						{
+							"listen": "test",
+							"script": {
+								"id": "179a918a-9d12-4d3a-9594-17818fefa228",
+								"type": "text/javascript",
+								"exec": [
+									"setTimeout(function(){}, 5000);",
+									"tests[\"Status code is 200\"] = responseCode.code === 200;"
+								]
+							}
+						}
+					],
+					"request": {
+						"method": "POST",
+						"header": [
+							{
+								"key": "Content-Type",
+								"value": "application/json"
+							}
+						],
+						"body": {
+							"mode": "raw",
+							"raw": "{\"controlOperation\":\"start\"}"
+						},
+						"url": {
+							"raw": "{{url}}/v1.0/jvms/1/commands?wait=TRUE",
+							"host": [
+								"{{url}}"
+							],
+							"path": [
+								"v1.0",
+								"jvms",
+								"1",
+								"commands"
+							],
+							"query": [
+								{
+									"key": "wait",
+									"value": "TRUE",
+									"equals": true
+								}
+							]
+						},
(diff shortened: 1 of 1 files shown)


jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -7182,6 +7182,128 @@
...
+								"key": "Content-Type",
+								"value": "application/json"
+							}
+						],
+						"body": {
+							"mode": "raw",
+							"raw": "{\"controlOperation\":\"start\"}"
+						},
+						"url": {
+							"raw": "{{url}}/v1.0/jvms/1/commands?wait=TRUE",
```
jmahmoodcc, line 7220: We can't assume that there will always be a JVM ID of 1 available. This needs to be a an environment variable that references a JVM created as part of the postman tests.

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -7182,6 +7182,128 @@
...
+							"host": [
+								"{{url}}"
+							],
+							"path": [
+								"v1.0",
+								"jvms",
+								"1",
```
jmahmoodcc, line 7227: Same thing here - we can't assume a JVM with an ID of 1 will always exist.

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -7182,6 +7182,128 @@
...
+				{
+					"name": "start jvm with wait true",
+					"event": [
+						{
+							"listen": "test",
+							"script": {
+								"id": "179a918a-9d12-4d3a-9594-17818fefa228",
+								"type": "text/javascript",
+								"exec": [
+									"setTimeout(function(){}, 5000);",
```
jmahmoodcc, line 7201: Do we still need this timeout in the test if the call to start the JVM is using setting synchronous flag to true?

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -7182,6 +7182,128 @@
...
+									"equals": true
+								}
+							]
+						},
+						"description": ""
+					},
+					"response": []
+				},
+				{
+					"name": "get health check application",
```
jmahmoodcc, line 7243: We need to remove all references to the health check application

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -7182,6 +7182,128 @@
...
+									"tests[\"Status code is 200\"] = responseCode.code === 200;"
+								]
+							}
+						}
+					],
+					"request": {
+						"method": "GET",
+						"header": [],
+						"body": {},
+						"url": "https://{{host}}/hct/",
```
jmahmoodcc, line 7260: We shouldn't be expecting the health check application to be deployed. I thought we had a simple application that post man deploys. 

Should these tests be in the health check postman test suite instead?

jmahmoodcc (changes requested):
Do we need to move these tests to the health check postman test suite?
The Jwala tests shouldn't have any reference to, or assume that, the CTP internal health check application is deployed.

jmahmoodcc:
Moving to a new collection in health check postman tests

Issue #206, opened 2018-05-21
Build failed with an exception


Ein-nor:
I am only a system administrator and I have no programming experience.

If I follow your building instructions I get the following error message:

FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':jwala-webapp:installBabelCli'.
> Process 'command 'C:\Users\Ron\Documents\git\jwala\jwala-webapp\build\nodejs\node-v0.11.10-windows-x64\bin\node.exe''
finished with non-zero exit value 1

I also tried to build jwala on linux an macosx.
Same result

jmahmoodcc:
@ZoeLeah can you try updating your repo now and try again?

PR #207, opened 2018-07-31
Fix for - Deleting JVMs in failed state throws an exception

Karthik263:
Fix for - Deleting JVMs in failed state throws an exception

### DIFF
--- a/jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
+++ b/jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
@@ -1118,6 +1118,20 @@ public class JvmServiceImplTest extends VerificationBehaviorSupport {
         verify(Config.mockJvmPersistenceService, never()).removeJvm(id);
     }
 
+
+    @Test
+    public void testDeleteFailedJvm() {
+        final CommandOutput mockCommandOutput = mock(CommandOutput.class);
+        when(mockJvm.getState()).thenReturn(JvmState.JVM_FAILED);
+        when(Config.mockJvmPersistenceService.getJvm(id)).thenReturn(mockJvm);
+        when(mockCommandOutput.getReturnCode()).thenReturn(new ExecReturnCode(0));
+        when(Config.mockJvmControlService.controlJvm(any(ControlJvmRequest.class), eq(user))).thenReturn(mockCommandOutput);
+        jvmService.deleteJvm(id, true, user);
+        verify(Config.mockJvmControlService).controlJvm(any(ControlJvmRequest.class), eq(user));
+        verify(Config.mockJvmPersistenceService).removeJvm(id);
+    }
+
+
     @Configuration
     static class Config {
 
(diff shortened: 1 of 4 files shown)


jwala-webapp/build.gradle
```diff
@@ -124,9 +124,9 @@ task watchJsx(type: Exec) {
 
 node {
     // Version of node to use.
-    version = '0.11.10'
+    version = '6.9.4'
 
-    npmVersion = '3.3.12'
+    npmVersion = '2.1.5'
```
rahulsayini: can you try with npmVersion 3.3.12

rahulsayini (approved):

jmahmoodcc (commented):
Just a general question: why are there 2 delete JVM methods? When does one get used instead of the other?

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
```diff
@@ -1118,6 +1118,20 @@ public void testDeleteNonStoppedAndNotNewJvm() {
         verify(Config.mockJvmPersistenceService, never()).removeJvm(id);
     }
 
+
+    @Test
+    public void testDeleteFailedJvm() {
```
jmahmoodcc, line 1123: Does the other delete method need a test? This method only tests the hardDelete

jwala-webapp/build.gradle
```diff
@@ -124,7 +124,7 @@ task watchJsx(type: Exec) {
 
 node {
     // Version of node to use.
-    version = '0.11.10'
+    version = '6.9.4'
```
jmahmoodcc, line 127: Just curious: can you explain why this stopped working if the version never changed before, and why do we have to jump from 0.11.10 all the way to 6.9.4?

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
Karthik263: There are actually three deleteJvm() methods.I updated the first two which belongs to Version 1 & 2 Service implementations. The third one is used internally and does not require this change.

Regarding Junits, I have done Junit for Version 1 implementation .For Version 2 service implementation ,I don't see existing  Junits /use cases .Hence, Skipped them for now.

jwala-webapp/build.gradle
Karthik263: There seems to be an issue with npm Version compatibility and we could see an issue(https://github.com/cerner/jwala/issues/206) reported on the same in May 2018 by one of the Jwala user. We did a lot of trial and error and finally able to fix this up with Version 6.9.4. This version was mentioned as a solution in one of the forums too. Not sure exactly on what went wrong on the NPM side.!

jmahmoodcc: Sounds good.

For future reference, add comments like this to the commit message in Git so that the Git log will contain the justification for the version update.

jmahmoodcc (approved):

PR #209, opened 2018-10-03
updated code to check and download upgraded JDK

Karthik263:
This pull request is related to JDK Upgrade enhancement.

### DIFF
--- a/jwala-services/src/main/java/com/cerner/jwala/service/jvm/JvmService.java
+++ b/jwala-services/src/main/java/com/cerner/jwala/service/jvm/JvmService.java
@@ -12,105 +12,119 @@ import com.cerner.jwala.service.jvm.impl.JvmHttpRequestResult;
 
 import java.util.List;
 
-
 public interface JvmService {
 
-    Jvm createJvm(CreateJvmAndAddToGroupsRequest createJvmAndAddToGroupsRequest, User user);
+	Jvm createJvm(CreateJvmAndAddToGroupsRequest createJvmAndAddToGroupsRequest, User user);
 
-    Jvm getJvm(final Identifier<Jvm> aJvmId);
+	Jvm getJvm(final Identifier<Jvm> aJvmId);
 
-    Jvm getJvm(final String jvmName);
+	Jvm getJvm(final String jvmName);
 
-    List<Jvm> getJvms();
+	List<Jvm> getJvms();
 
-    Jvm updateJvm(final UpdateJvmRequest updateJvmRequest, boolean updateJvmPassword);
+	Jvm updateJvm(final UpdateJvmRequest updateJvmRequest, boolean updateJvmPassword);
 
-    void deleteJvm(Identifier<Jvm> id, boolean hardDelete, User user);
+	void deleteJvm(Identifier<Jvm> id, boolean hardDelete, User user);
 
+	Jvm generateAndDeployJvm(String jvmName, User user);
 
+	Jvm generateAndDeployFile(String jvmName, String fileName, User user);
 
-    Jvm generateAndDeployJvm(String jvmName, User user);
+	void performDiagnosis(Identifier<Jvm> aJvmId, User user);
 
-    Jvm generateAndDeployFile(String jvmName, String fileName, User user);
+	List<String> getResourceTemplateNames(final String jvmName);
 
-    void performDiagnosis(Identifier<Jvm> aJvmId, User user);
+	String getResourceTemplate(final String jvmName, final String resourceTemplateName, final boolean tokensReplaced);
 
-    List<String> getResourceTemplateNames(final String jvmName);
+	String updateResourceTemplate(final String jvmName, final String resourceTemplateName, final String template);
 
-    String getResourceTemplate(final String jvmName, final String resourceTemplateName, final boolean tokensReplaced);
+	String previewResourceTemplate(String fileName, String jvmName, String groupName, String template);
 
-    String updateResourceTemplate(final String jvmName, final String resourceTemplateName, final String template);
+	/**
+	 * Update state of JVM
+	 * 
+	 * @param id    JVM id
+	 * @param state JVM State
+	 */
+	public void updateState(Identifier<Jvm> id, JvmState state);
 
-    String previewResourceTemplate(String fileName, String jvmName, String groupName, String template);
+	/**
+	 * Ping's the JVM and updates its state.
+	 *
+	 * @param jvm the JVM
+	 */
+	JvmHttpRequestResult pingAndUpdateJvmState(Jvm jvm);
 
-    /**
-     * Update state of JVM
-     * @param id JVM id
-     * @param state JVM State
-     */
-    public void updateState(Identifier<Jvm> id, JvmState state);
+	/**
+	 * Deploy application context xml for JVM's
+	 * 
+	 * @param jvm  JVM
+	 * @param user User
+	 */
+	void deployApplicationContextXMLs(Jvm jvm, User user);
 
-    /**
-     * Ping's the JVM and updates its state.
-     *
-     * @param jvm the JVM
-     */
-    JvmHttpRequestResult pingAndUpdateJvmState(Jvm jvm);
+	/**
+	 * get Get the count of JVM's which are started
+	 * 
+	 * @param groupName
+	 * @return
+	 */
+	public Long getJvmStartedCount(String groupName);
 
-    /**
-     * Deploy application context xml for JVM's
-     * @param jvm JVM
-     * @param user User
-     */
-    void deployApplicationContextXMLs(Jvm jvm, User user);
+	/**
+	 * Get the jvm count for a group
+	 * 
+	 * @param groupName
+	 * @return
+	 */
+	public Long getJvmCount(String groupName);
 
-    /**
-     * get Get the count of JVM's which are started
-     * @param groupName
-     * @return
-     */
-    public Long getJvmStartedCount(String groupName);
+	/**
+	 * Get the count of stopped JVM's
+	 * 
+	 * @param groupName
+	 * @return
+	 */
+	public Long getJvmStoppedCount(String groupName);
 
-    /**
-     * Get the jvm count for a group
-     * @param groupName
-     * @return
-     */
-    public Long getJvmCount(String groupName);
+	/**
+	 * Get the count of forcibly stopped JVM's
+	 * 
+	 * @param groupName
+	 * @return
+	 */
+	public Long getJvmForciblyStoppedCount(String groupName);
 
-    /**
-     * Get the count of stopped JVM's
-     * @param groupName
-     * @return
-     */
-    public Long getJvmStoppedCount(String groupName);
+	/**
+	 * Create JVM default templates.
+	 *
+	 * @param jvmName
+	 * @param parentGroup
+	 */
+	void createDefaultTemplates(String jvmName, Group parentGroup);
 
-    /**
-     * Get the count of forcibly stopped JVM's
-     * @param groupName
-     * @return
-     */
-    public Long getJvmForciblyStoppedCount(String groupName);
+	/**
+	 * Delete a JVM
+	 * 
+	 * @param name     the name of the JVM to delete
+	 * @param userName
+	 */
+	void deleteJvm(String name, String userName);
 
-    /**
-     * Create JVM default templates.
-     *
-     * @param jvmName
-     * @param parentGroup
-     */
-    void createDefaultTemplates(String jvmName, Group parentGroup);
+	/**
+	 * Return the JVMs belonging to a group
+	 * 
+	 * @param name the name of the group of the JVMs
+	 * @return a collection of JVMs belonging to the group
+	 */
+	List<Jvm> getJvmsByGroupName(String name);
 
-    /**
-     * Delete a JVM
-     * @param name the name of the JVM to delete
-     * @param userName
-     */
-    void deleteJvm(String name, String userName);
-
-    /**
-     * Return the JVMs belonging to a group
-     * @param name the name of the group of the JVMs
-     * @return a collection of JVMs belonging to the group
-     */
-    List<Jvm> getJvmsByGroupName(String name);
+	/**
+	 * Upgrades the JDK of the mentioned JVM
+	 * 
+	 * @param jvmName
+	 * @param user
+	 * @return
+	 */
+	Jvm upgradeJDKAndDeployJvm(String jvmName, User user);
 }
--- a/jwala-services/src/main/java/com/cerner/jwala/service/jvm/JvmService.java
+++ b/jwala-services/src/main/java/com/cerner/jwala/service/jvm/JvmService.java
@@ -126,5 +126,5 @@ public interface JvmService {
 	 * @param user
 	 * @return
 	 */
-	Jvm upgradeJDKAndDeployJvm(String jvmName, User user);
+	Jvm upgradeJDK(String jvmName, User user);
 }
--- a/jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
+++ b/jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
@@ -25,6 +25,7 @@ import javax.persistence.NoResultException;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.SystemUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.tika.mime.MediaType;
 import org.joda.time.DateTime;
@@ -89,7 +90,8 @@ import com.cerner.jwala.service.webserver.component.ClientFactoryHelper;
 public class JvmServiceImpl implements JvmService {
     private static final Logger LOGGER = LoggerFactory.getLogger(JvmServiceImpl.class);
     private static final String MEDIA_TYPE_TEXT = "text";
-    private static final String RESOURCE_FILE_SETENV = "setenv";
+    private static final String RESOURCE_FILE_SETENV_WINDOWS = "setenv.bat";
+    private static final String RESOURCE_FILE_SETENV_LINUX = "setenv.sh";
     private final BinaryDistributionLockManager binaryDistributionLockManager;
     private final String topicServerStates;
     private final JvmPersistenceService jvmPersistenceService;
@@ -752,20 +754,25 @@ public class JvmServiceImpl implements JvmService {
 
 			// Step 1: Check if JDK Binaries exists
 			checkForJvmBinaries(jvm);
+
 			// Step 2: Distribute JDK binaries if required
 			distributeBinaries(jvm);
 
-			// Step 3: delete the service
+			// Step 3: Delete the service
 			deleteJvmService(jvm, user);
 
-			// Step 4: update the Setenv file with upgraded JDK
-			deployJvmResourceFilesForJDKUpgrade(jvm, user);
+			// Step 4: Generate and deploy Setenv file with upgraded JDK
+			if (SystemUtils.IS_OS_WINDOWS) {
+				generateAndDeployFile(jvmName, RESOURCE_FILE_SETENV_WINDOWS, user);
+			} else {
+				generateAndDeployFile(jvmName, RESOURCE_FILE_SETENV_LINUX, user);
+			}
 
-			// Step 5: re-install the service
+			// Step 5: Re-install the service
 			installJvmWindowsService(jvm, user);
 
 			didSucceed = true;
-		} catch (CommandFailureException | IOException e) {
+		} catch (CommandFailureException e) {
 			LOGGER.error("Failed to upgrade JDK :: the JVM config for {}", jvm.getJvmName(), e);
 			throw new InternalErrorException(FaultType.REMOTE_COMMAND_FAILURE,
 					"Failed to generate the JVM config: " + jvm.getJvmName(), e);
@@ -784,7 +791,7 @@ public class JvmServiceImpl implements JvmService {
 		}
 		return jvm;
 	}
-    
+
     String generateJvmConfigJar(Jvm jvm) throws CommandFailureException {
         long startTime = System.currentTimeMillis();
         LOGGER.debug("Start generateJvmConfigJar ");
@@ -862,78 +869,7 @@ public class JvmServiceImpl implements JvmService {
             }
         }
     }
-    
-	/**
-	 * This method modifies the setenv file with the JDK home location during JDK
-	 * upgrade process
-	 * 
-	 * @param jvm
-	 * @param user
-	 * @throws IOException
-	 * @throws CommandFailureException
-	 */
-	private void deployJvmResourceFilesForJDKUpgrade(Jvm jvm, User user) throws IOException, CommandFailureException {
-		final Map<String, ScpDestination> generatedFiles = generateResourceFilesByName(jvm.getJvmName(),
-				RESOURCE_FILE_SETENV);
-		if (generatedFiles != null) {
-			for (Map.Entry<String, ScpDestination> entry : generatedFiles.entrySet()) {
-				final ScpDestination scpDestination = entry.getValue();
-				String sourceFile = entry.getKey();
-				secureCopyFileToJvm(jvm, sourceFile, scpDestination.destPath, user, scpDestination.overwrite);
 
-			}
-		}
-	}
-	
-	/**
-	 * This method generates the targeted resource file passed as resourceFileName
-	 * argument.
-	 * 
-	 * @param jvmName
-	 * @param resourceFileName
-	 * @return Map<String,ScpDestination>
-	 * @throws IOException
-	 * 
-	 */
-	private Map<String, ScpDestination> generateResourceFilesByName(final String jvmName, String resourceFileName)
-			throws IOException {
-		Map<String, ScpDestination> generatedFiles = new HashMap<>();
-		final List<JpaJvmConfigTemplate> jpaJvmConfigTemplateList = jvmPersistenceService.getConfigTemplates(jvmName);
-
-		for (final JpaJvmConfigTemplate jpaJvmConfigTemplate : jpaJvmConfigTemplateList) {
-			// Generate only the file mentioned as resourceFileName
-			if (!jpaJvmConfigTemplate.getTemplateName().contains(resourceFileName)) {
-				continue;
-			}
-			final ResourceGroup resourceGroup = resourceService.generateResourceGroup();
-			final Jvm jvm = jvmPersistenceService.findJvmByExactName(jvmName);
-
-			String resourceTemplateMetaDataString = "";
-			resourceTemplateMetaDataString = resourceService.generateResourceFile(
-					jpaJvmConfigTemplate.getTemplateName(), jpaJvmConfigTemplate.getMetaData(), resourceGroup, jvm,
-					ResourceGeneratorType.METADATA);
-			final ResourceTemplateMetaData resourceTemplateMetaData = resourceService
-					.getMetaData(resourceTemplateMetaDataString);
-			final String deployFileName = resourceTemplateMetaData.getDeployFileName();
-			if (resourceTemplateMetaData.getContentType().getType().equalsIgnoreCase(MEDIA_TYPE_TEXT)
-					|| MediaType.APPLICATION_XML.equals(resourceTemplateMetaData.getContentType())) {
-				final String generatedResourceStr = resourceService.generateResourceFile(
-						jpaJvmConfigTemplate.getTemplateName(), jpaJvmConfigTemplate.getTemplateContent(),
-						resourceGroup, jvm, ResourceGeneratorType.TEMPLATE);
-				generatedFiles.put(
-						createConfigFile(
-								ApplicationProperties.get(PropertyKeys.PATHS_GENERATED_RESOURCE_DIR) + '/' + jvmName,
-								deployFileName, generatedResourceStr),
-						new ScpDestination(resourceTemplateMetaData.getDeployPath() + '/' + deployFileName,
-								resourceTemplateMetaData.isOverwrite()));
-			} else {
-				generatedFiles.put(jpaJvmConfigTemplate.getTemplateContent(),
-						new ScpDestination(resourceTemplateMetaData.getDeployPath() + '/' + deployFileName,
-								resourceTemplateMetaData.isOverwrite()));
-			}
-		}
-		return generatedFiles;
-	}
 	 
     private void installJvmWindowsService(Jvm jvm, User user) {
         ControlJvmRequest controlJvmRequest = ControlJvmRequestFactory.create(JvmControlOperation.INSTALL_SERVICE, jvm);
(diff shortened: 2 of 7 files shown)


jwala-services/src/main/java/com/cerner/jwala/service/jvm/JvmService.java
```diff
@@ -12,105 +12,119 @@
...
+	List<Jvm> getJvmsByGroupName(String name);
+
+	/**
+	 * Upgrades the JDK of the mentioned JVM
+	 * 
+	 * @param jvmName
+	 * @param user
+	 * @return
+	 */
+	Jvm upgradeJDKAndDeployJvm(String jvmName, User user);
```
Karthik263: This is the only line added in this file . Rest are formatting changes.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +704,219 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
...
+    }
+
+    /**
+     * This method will deploy necessary scripts to Jwala Directory during JDK upgrade process
+     * @param jvm
+     * @param user
+     * @throws CommandFailureException
+     * @throws IOException
+     */
+    protected void deployScriptsToUserJwalaScriptsDir_JDKUpgrade(Jvm jvm, User user) throws CommandFailureException, IOException {
```
Karthik263: This method .. I just created for next step.. For now we are not using this method.We only use upgradeJDKAnd DeployJVM method

jmahmoodcc:
@Karthik263, please get in the habit of editing the description for the pull request. It helps provide context when other developers get involved on the project.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/JvmService.java
jmahmoodcc: Please fix your IDE to not reformat the files and make sure the commits are clean.
This file has gone through many revisions without formatting issues - as you can see, applying your custom formatting makes the review process more difficult.

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/jvm/JvmServiceRest.java
```diff
@@ -178,4 +178,21 @@ Response previewResourceTemplate(@ApiParam(value = "The name of the JVM", requir
...
                                      @ApiParam(value = "The content to be generated", required = true) String template);
 
+    /**
+     * Upgraded the JDK version of the mentioned JVM.
+     * @param jvmName the name of the JVM
+     * @param aUser the user
+     * @return {@link Response}
+     */
+    @PUT
+    @Path("/{jvmName}/upgradeJDK")
```
jmahmoodcc: For REST APIs nouns are preferred, and the HTTP action implies the behavior of the call.
In this instance the REST API can simply be /{jvmName}/jdk, and because it's an HTTP POST then it's implied that the JDK for the JVM is being updated.

jwala-webservices/src/main/java/com/cerner/jwala/ws/rest/v1/service/jvm/JvmServiceRest.java
```diff
@@ -178,4 +178,21 @@ Response previewResourceTemplate(@ApiParam(value = "The name of the JVM", requir
...
+    @ApiOperation(value = "upgrade the JDK version for the JVM",
+            response = Jvm.class
+    )
+    @ApiResponses(@ApiResponse(code = 500, message = "Remote upgradation of JDK failed"))
```
jmahmoodcc: This is a very minor thing, but please don't use the word "upgradation" :)
You can just say something like "Attempt to upgrade the JDK failed."

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
jmahmoodcc: I know all of this code is going to be changed eventually, but try to follow the Java conventions even when you're writing experimental code. It will make clean up easier in the end.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/JvmService.java
Karthik263: Sure... My eclipse was configured with CAC eclipse formatter. Just removed that for CTP projects.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -86,6 +103,9 @@
     private final HistoryFacadeService historyFacadeService;
     private final BinaryDistributionService binaryDistributionService;
     private final FileUtility fileUtility;
+    //JDK Upgrade Resource file names Array
+   private static String[] jdk_upgrade_resourceNamesArray= {"setenv"};
```
jmahmoodcc: Formatting: fix the spacing and the name of the array doesn't match Java standards

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +724,193 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
         }
 
     }
+    
+    public Jvm upgradeJDKAndDeployJvm(String jvmName, User user) {
+        boolean didSucceed = false;
+        Jvm jvm = getJvm(jvmName);
+        LOGGER.debug("Start upgradeJDKAndDeployJvm for {} by user {}", jvmName, user.getId());
```
jmahmoodcc: Change this to LOGGER.info, we want to see all the actions initiated by the user in the logs

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +724,193 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
...
+        historyFacadeService.write(jvm.getHostName(), jvm.getGroups(), "Starting to upgrade JDK :: JVM " +
+                jvm.getJvmName() + " on host " + jvm.getHostName(), EventType.USER_ACTION_INFO, user.getId());
+
+        //add write lock for multiple write
+        binaryDistributionLockManager.writeLock(jvmName + "-" + jvm.getId().toString());
+
+        try {
+            if (jvm.getState().isStartedState()) {
+                final String errorMessage = "The remote JVM " + jvm.getJvmName() + " must be stopped before attempting to generate the JVM";
+                LOGGER.info(errorMessage);
```
jmahmoodcc: Make this LOGGER.error

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +724,193 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
...
+    public Jvm upgradeJDKAndDeployJvm(String jvmName, User user) {
+        boolean didSucceed = false;
+        Jvm jvm = getJvm(jvmName);
+        LOGGER.debug("Start upgradeJDKAndDeployJvm for {} by user {}", jvmName, user.getId());
+
+        historyFacadeService.write(jvm.getHostName(), jvm.getGroups(), "Starting to upgrade JDK :: JVM " +
+                jvm.getJvmName() + " on host " + jvm.getHostName(), EventType.USER_ACTION_INFO, user.getId());
+
+        //add write lock for multiple write
+        binaryDistributionLockManager.writeLock(jvmName + "-" + jvm.getId().toString());
```
jmahmoodcc: Create a final variable for the lock name and reuse it for unlocking

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +724,193 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
...
+
+        try {
+            if (jvm.getState().isStartedState()) {
+                final String errorMessage = "The remote JVM " + jvm.getJvmName() + " must be stopped before attempting to generate the JVM";
+                LOGGER.info(errorMessage);
+                throw new InternalErrorException(FaultType.REMOTE_COMMAND_FAILURE, errorMessage);
+            }
+            //Does this required??
+            validateJvmAndAppResources(jvm);
```
jmahmoodcc: Remove this line. It only validates application resources for the JVM and this method should only deploy the new JDK

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +724,193 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
...
+        LOGGER.debug("Start upgradeJDKAndDeployJvm for {} by user {}", jvmName, user.getId());
+
+        historyFacadeService.write(jvm.getHostName(), jvm.getGroups(), "Starting to upgrade JDK :: JVM " +
+                jvm.getJvmName() + " on host " + jvm.getHostName(), EventType.USER_ACTION_INFO, user.getId());
+
+        //add write lock for multiple write
+        binaryDistributionLockManager.writeLock(jvmName + "-" + jvm.getId().toString());
+
+        try {
+            if (jvm.getState().isStartedState()) {
```
jmahmoodcc: Check for the JVM not in NEW state as well. We want to make sure that the JVM was already generated and deployed as a service. This method is meant to only update the JDK of an existing JVM service.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +724,193 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
...
+            }
+            //Does this required??
+            validateJvmAndAppResources(jvm);
+
+            checkForJvmBinaries(jvm);
+
+            distributeBinaries(jvm);
+            
+            //This will update the Setenv file with upgraded JDK
+            deployJvmResourceFilesForJDKUpgrade(jvm, user);
```
jmahmoodcc: This deploys all of the resources for the JVM. You only want to deploy the setenv.bat or setenv.sh that defines the JAVA_HOME.

Also, when does that file get updated? I would expect it to be updated as part of this method before it is deployed.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +724,193 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
...
+            // set the state to stopped
+            updateState(jvm.getId(), JvmState.JVM_STOPPED);
+
+            didSucceed = true;
+        } catch (CommandFailureException | IOException e) {
+            LOGGER.error("Failed to upgrade JDK :: the JVM config for {}", jvm.getJvmName(), e);
+            throw new InternalErrorException(FaultType.REMOTE_COMMAND_FAILURE, "Failed to generate the JVM config: " + jvm.getJvmName(), e);
+        } finally {
+            binaryDistributionLockManager.writeUnlock(jvmName + "-" + jvm.getId().toString());
+            LOGGER.debug("End upgrade JDK for {} by user {}", jvmName, user.getId());
```
jmahmoodcc: Make this LOGGER.info so we can see that end of the upgrade in the log.

jmahmoodcc (commented):
It looks like this code is still unfinished. The method for updating the setenv file isn't used, and it doesn't look to be functioning anyway.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
Karthik263: This is fixed in the latest check-in.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
Karthik263: I have re-used the same existing mechanism for lock name. I have not created any new variable.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
Karthik263: This is fixed in the latest check-in.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
Karthik263: This is fixed in the latest check-in.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
Karthik263: This is fixed in the latest check-in.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
Karthik263: deployJvmResourceFilesForJDKUpgrade() is a newly created method. The setenv file update is handled in this method

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
Karthik263: This is fixed in the latest check-in.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +721,67 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
         }
 
     }
-
+    
+   /**
```
jmahmoodcc: If you're adding a javadoc comment please make sure to at least document all the parameters and the the returned value. You should be able to setup your IDE to create a template for you.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +721,67 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
         }
 
     }
-
+    
+   /**
+    * This method upgrades the JDK version for the mentioned JVM
+    * 
+    */
+	public Jvm upgradeJDKAndDeployJvm(String jvmName, User user) {
```
jmahmoodcc: Another minor formatting issue: please setup your IDE to use 4 spaces when tabbing - all of the code should be uniformly formatted

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -774,15 +851,35 @@ private void deployJvmConfigJar(Jvm jvm, User user, String jvmJar) throws Comman
...
             for (Map.Entry<String, ScpDestination> entry : generatedFiles.entrySet()) {
                 final ScpDestination scpDestination = entry.getValue();
                 secureCopyFileToJvm(jvm, entry.getKey(), scpDestination.destPath, user, scpDestination.overwrite);
             }
         }
     }
-
+    
+/**
+ * This method modifies the setenv file with the JDK home location during JDK upgrade process
```
jmahmoodcc: Please fill in the rest of this template

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -774,15 +851,35 @@ private void deployJvmConfigJar(Jvm jvm, User user, String jvmJar) throws Comman
...
+ * @param jvm
+ * @param user
+ * @throws IOException
+ * @throws CommandFailureException
+ */
+	private void deployJvmResourceFilesForJDKUpgrade(Jvm jvm, User user) throws IOException, CommandFailureException {
```
jmahmoodcc: Formatting: tabs = 4 spaces

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -774,15 +851,35 @@ private void deployJvmConfigJar(Jvm jvm, User user, String jvmJar) throws Comman
...
+		final Map<String, ScpDestination> generatedFiles = generateResourceFiles(jvm.getJvmName(), true);
```
jmahmoodcc: You don't need to generate all of the resource files if you're only doing it for the setenv.bat/sh resource. I know you changed this method to only work on the setenv resource, but the implication from reading the code is that multiple files are being generated. Also, filtering the loop inside this method using a stateful variable is another issue that I'll address later.

There are a few methods available to get a single resource by it's name and JVM name (there should really only be one way to do it, but that's another problem outside the scope of this review). Use one of those methods in this case.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -774,15 +851,35 @@ private void deployJvmConfigJar(Jvm jvm, User user, String jvmJar) throws Comman
...
+		if (generatedFiles != null) {
+			for (Map.Entry<String, ScpDestination> entry : generatedFiles.entrySet()) {
+				final ScpDestination scpDestination = entry.getValue();
+				String sourceFile = entry.getKey();
+				if (sourceFile.contains(RESOURCE_FILE_SETENV)) {
+					secureCopyFileToJvm(jvm, sourceFile, scpDestination.destPath, user, scpDestination.overwrite);
```
jmahmoodcc: You wouldn't need to iterate through the generatedFiles list or add a filter here if you only generated the resource you needed.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
jmahmoodcc: Also rename the method to just upgradeJDK.
This method isn't really deploying the JVM, but rather deleting and recreating the service for the JVM. Add those details in the javadoc for the description of the method.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -774,15 +851,35 @@ private void deployJvmConfigJar(Jvm jvm, User user, String jvmJar) throws Comman
     }
 
     private void deployJvmResourceFiles(Jvm jvm, User user) throws IOException, CommandFailureException {
-        final Map<String, ScpDestination> generatedFiles = generateResourceFiles(jvm.getJvmName());
+        final Map<String, ScpDestination> generatedFiles = generateResourceFiles(jvm.getJvmName(),false);
```
jmahmoodcc: Remove this parameter, see my comments below

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -1013,10 +1110,16 @@ public Long getJvmForciblyStoppedCount(final String groupName) {
         return jvmPersistenceService.getJvmForciblyStoppedCount(groupName);
     }
 
-    private Map<String, ScpDestination> generateResourceFiles(final String jvmName) throws IOException {
+    private Map<String, ScpDestination> generateResourceFiles(final String jvmName,boolean isJDKUpgradeService) throws IOException {
```
jmahmoodcc: Remove this parameter to keep this method generic.
In terms of generating resource files we don't need to add filtering or special processing. There are more efficient ways to do this. If you need to, refactor out the piece of this code that generates the resource you need and call that method.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -1013,10 +1110,16 @@ public Long getJvmForciblyStoppedCount(final String groupName) {
...
 
-    private Map<String, ScpDestination> generateResourceFiles(final String jvmName) throws IOException {
+    private Map<String, ScpDestination> generateResourceFiles(final String jvmName,boolean isJDKUpgradeService) throws IOException {
         Map<String, ScpDestination> generatedFiles = new HashMap<>();
         final List<JpaJvmConfigTemplate> jpaJvmConfigTemplateList = jvmPersistenceService.getConfigTemplates(jvmName);
+      
         for (final JpaJvmConfigTemplate jpaJvmConfigTemplate : jpaJvmConfigTemplateList) {
+			// If isJDKUpgradeService is true ,then proceed with generating Resource files
+			// only setenv file.
+			if (isJDKUpgradeService && !jpaJvmConfigTemplate.getTemplateName().contains(RESOURCE_FILE_SETENV)) {
```
jmahmoodcc: Remove these changes. See my previous comments.

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
```diff
@@ -1281,5 +1281,56 @@ public JvmService getJvmService() {
...
+        templateNames.add("setenv.bat");
+        when(Config.mockJvmPersistenceService.getResourceTemplateNames(anyString())).thenReturn(templateNames);
+
+        Jvm response = jvmService.upgradeJDKAndDeployJvm(mockJvm.getJvmName(), Config.mockUser);
+        assertEquals(response.getJvmName(), mockJvm.getJvmName());
+
+        when(mockJvm.getState()).thenReturn(JvmState.JVM_NEW);
+        boolean exceptionThrown = false;
+        try {
+            jvmService.upgradeJDKAndDeployJvm(mockJvm.getJvmName(), Config.mockUser);
```
jmahmoodcc: Make this a separate test. It should be easy because you won't need all the mock setup from before.

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
```diff
@@ -1281,5 +1281,56 @@ public JvmService getJvmService() {
...
+        } catch (Exception e) {
+            exceptionThrown = true;
```
jmahmoodcc: When you make this a separate test you can use the `@Test(expected = Exception.class` to test for exceptions being thrown instead of setting and checking a flag.

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
```diff
@@ -1281,5 +1281,56 @@ public JvmService getJvmService() {
         }
 
     }
+    
+  
+    @Test
+    public void testUpgradeJDKAndDeployJvm() throws CommandFailureException, IOException {
```
jmahmoodcc: Move this test up with the rest of the tests instead of at the end after all of the test configuration mock objects.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +721,67 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
...
+			LOGGER.error("Failed to upgrade JDK :: the JVM config for {}", jvm.getJvmName(), e);
+			throw new InternalErrorException(FaultType.REMOTE_COMMAND_FAILURE,
+					"Failed to generate the JVM config: " + jvm.getJvmName(), e);
+		} finally {
+			binaryDistributionLockManager.writeUnlock(jvmName + "-" + jvm.getId().toString());
+			LOGGER.info("End upgrade JDK for {} by user {}", jvmName, user.getId());
+
+			final EventType eventType = didSucceed ? EventType.SYSTEM_INFO : EventType.SYSTEM_ERROR;
+
+			String historyMessage = didSucceed
```
jmahmoodcc, line 788: When the method succeeds add the JDK version that was installed to the history message as well.

jmahmoodcc (changes requested):
Clean up and refactoring required.
See my inline comments.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,7 +721,70 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
         }
 
     }
-
+    
+	/**
+	 * This method upgrades the JDK version for the mentioned JVM
+	 * 
+	 * @param jvmName
```
jmahmoodcc, line 730: Fill in the descriptions for these parameters and return values

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -782,7 +862,79 @@ private void deployJvmResourceFiles(Jvm jvm, User user) throws IOException, Comm
             }
         }
     }
-
+    
+	/**
+	 * This method modifies the setenv file with the JDK home location during JDK
```
jmahmoodcc: This description is inaccurate as it implies that the setenv template itself would be modified. The description should state that the setenv.bat or .sh is being generated again in order to pick up the new JDK version.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -782,7 +862,79 @@ private void deployJvmResourceFiles(Jvm jvm, User user) throws IOException, Comm
             }
         }
     }
-
+    
+	/**
+	 * This method modifies the setenv file with the JDK home location during JDK
+	 * upgrade process
+	 * 
+	 * @param jvm
```
jmahmoodcc: Fill in the description for all of these parameters, exceptions, and return types

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -782,7 +862,79 @@ private void deployJvmResourceFiles(Jvm jvm, User user) throws IOException, Comm
...
+	 * This method modifies the setenv file with the JDK home location during JDK
+	 * upgrade process
+	 * 
+	 * @param jvm
+	 * @param user
+	 * @throws IOException
+	 * @throws CommandFailureException
+	 */
+	private void deployJvmResourceFilesForJDKUpgrade(Jvm jvm, User user) throws IOException, CommandFailureException {
+		final Map<String, ScpDestination> generatedFiles = generateResourceFilesByName(jvm.getJvmName(),
```
jmahmoodcc: This method already exists ... you only needed to find the name of the setenv file (either.bat or .sh) and then call the existing method (https://github.com/cerner/jwala/blob/JDK_upgrade_enhancement/jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java#L981)

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -782,7 +862,79 @@ private void deployJvmResourceFiles(Jvm jvm, User user) throws IOException, Comm
...
+				RESOURCE_FILE_SETENV);
+		if (generatedFiles != null) {
+			for (Map.Entry<String, ScpDestination> entry : generatedFiles.entrySet()) {
+				final ScpDestination scpDestination = entry.getValue();
+				String sourceFile = entry.getKey();
+				secureCopyFileToJvm(jvm, sourceFile, scpDestination.destPath, user, scpDestination.overwrite);
```
jmahmoodcc: The method mentioned above takes care of copying the file as well so there is no need to do this either

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -782,7 +862,79 @@ private void deployJvmResourceFiles(Jvm jvm, User user) throws IOException, Comm
...
+	 * This method generates the targeted resource file passed as resourceFileName
+	 * argument.
+	 * 
+	 * @param jvmName
+	 * @param resourceFileName
+	 * @return Map<String,ScpDestination>
+	 * @throws IOException
+	 * 
+	 */
+	private Map<String, ScpDestination> generateResourceFilesByName(final String jvmName, String resourceFileName)
```
jmahmoodcc: Remove this method as it is not needed

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -1016,8 +1168,9 @@ public Long getJvmForciblyStoppedCount(final String groupName) {
     private Map<String, ScpDestination> generateResourceFiles(final String jvmName) throws IOException {
         Map<String, ScpDestination> generatedFiles = new HashMap<>();
         final List<JpaJvmConfigTemplate> jpaJvmConfigTemplateList = jvmPersistenceService.getConfigTemplates(jvmName);
+      
         for (final JpaJvmConfigTemplate jpaJvmConfigTemplate : jpaJvmConfigTemplateList) {
-            final ResourceGroup resourceGroup = resourceService.generateResourceGroup();
+			final ResourceGroup resourceGroup = resourceService.generateResourceGroup();
```
jmahmoodcc, line 1111: Fix formatting

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
```diff
@@ -1132,6 +1132,74 @@ public void testDeleteFailedJvm() {
     }
 
 
+	@Test(expected = Exception.class)
```
jmahmoodcc: Try to be as specific as possible when using expected.
Catching Exception will mask any test failures that aren't specific to your test which could result in false positives. In this case I believe you want InternalErrorException which is a Jwala specific exception type.

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
```diff
@@ -1132,6 +1132,74 @@ public void testDeleteFailedJvm() {
...
+		when(Config.mockResourceService.generateResourceFile(anyString(), anyString(), any(ResourceGroup.class),
+				anyObject(), any(ResourceGeneratorType.class))).thenReturn("<server>some xml</server>");
+
+		final List<String> templateNames = new ArrayList<>();
+		templateNames.add("setenv.bat");
+		when(Config.mockJvmPersistenceService.getResourceTemplateNames(anyString())).thenReturn(templateNames);
+
+		Jvm response = jvmService.upgradeJDK(mockJvm.getJvmName(), Config.mockUser);
+		assertEquals(response.getJvmName(), mockJvm.getJvmName());
+
```
jmahmoodcc, line 1202: Add verify() calls to check that the mocks are getting called as expected and that the interactions with the other services are occurring as you expected.

Example verify calls: https://github.com/cerner/jwala/blob/JDK_upgrade_enhancement/jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java#L137

jmahmoodcc (changes requested):
It's important that we reuse as much of the existing methods as possible. Duplicating behavior results in extra maintenance, increases the complexity, and creates the potential for unnecessary bugs.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -1165,34 +1101,38 @@ public Long getJvmForciblyStoppedCount(final String groupName) {
...
+			} else {
+				generatedFiles.put(jpaJvmConfigTemplate.getTemplateContent(),
+						new ScpDestination(resourceTemplateMetaData.getDeployPath() + '/' + deployFileName,
+								resourceTemplateMetaData.isOverwrite()));
+			}
+		}
+		return generatedFiles;
+	}
 
     /**
```
Karthik263, line 1140: This unchanged method got formatted .!

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
jmahmoodcc: Please fill in this javadoc

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,6 +723,74 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
...
+			checkForJvmBinaries(jvm);
+
+			// Step 2: Distribute JDK binaries if required
+			distributeBinaries(jvm);
+
+			// Step 3: Delete the service
+			deleteJvmService(jvm, user);
+
+			// Step 4: Generate and deploy Setenv file with upgraded JDK
+			if (SystemUtils.IS_OS_WINDOWS) {
```
jmahmoodcc: This if-block should be used to set the resource name variable, and not used to branch the calling of the generateAndDeploy method.

Only call the method once, and have the variables be manipulated by a conditional. This way if the method API changes you only have to change it in one place.

`generatedAndDeployFile(jvmName, System.Utils.IS_OS_WINDOWS ? RESOURCE_FILE_SETENV_WINDOWS : RESOURCE_FILE_SETENV_LINUX, user);`

jmahmoodcc (changes requested):
Close, just some minor changes needed

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,6 +723,77 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
...
+			// Step 1: Check if JDK Binaries exists
+			checkForJvmBinaries(jvm);
+
+			// Step 2: Distribute JDK binaries if required
+			distributeBinaries(jvm);
+
+			// Step 3: Delete the service
+			deleteJvmService(jvm, user);
+
+			try {
```
jmahmoodcc: Don't nest try-catch blocks. 
The code is already wrapped in a try block, just add the NoResultException to the existing catch block as a handled exception.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -704,6 +723,77 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
...
+				// Step 4: Generate and deploy Setenv file with upgraded JDK
+				generateAndDeployFile(jvmName, resourceFileName, user);
+			} catch (NoResultException e) {
+				String errorMessage = resourceFileName + " resource file does not exist for JVM: " + jvm.getJvmName();
```
jmahmoodcc: I would be more explicit about the context of this error message. 
Now we know the resource doesn't exist, but why is that causing a failure? What was the expectation that the setenv resource needed to be available?

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
jmahmoodcc: Please fill in this javadoc, it will get flagged by static analysis tools as incomplete. The IDE should be highlighting this already, and it's an easy fix that provides long term benefits when done correctly.

jmahmoodcc (changes requested):
See comments in JvmServiceImpl

rahulsayini (approved):

PR #210, opened 2018-11-29
Jdk upgrade enhancement

Karthik263:
1. Streamlined the postman scripts to new Sub-division called "JDK Upgrade operations".Updated the Postman scripts to include a positive and negative workflow.
2. Added postman script to update the "JVM Values" before calling the upgrade scenario.
3. Removed hardcoded JVM name in the postman testcase.
4. Added Javadoc param description for upgradeJDK() method.


jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6256,96 +6257,112 @@
...
+{
+					"name": "upgrade JDK- Negative Test case",
 					"event": [
 						{
 							"listen": "test",
 							"script": {
 								"type": "text/javascript",
 								"exec": [
-									"tests[\"Status code is 201\"] = responseCode.code === 201;"
+									"tests[\"Status code is 500\"] = responseCode.code === 500;"
```
jmahmoodcc: Add a check for the actual error message itself.
Since multiple errors could be returned we want to verify that the message is relevant and expressive so that debugging is easier.

Here is a sample from the external properties test:

 "name": "download external properties error No external properties uploaded",
                                        "event": [
                                                {
                                                        "listen": "test",
                                                        "script": {
                                                                "type": "text/javascript",
                                                                "exec": [
                                                                        "tests[\"Status code is 500\"] = responseCode.code === 500;",
                                                                        "",
                                                                        "var jsonData = JSON.parse(responseBody);",
                                                                        "tests[\"Assert No file error\"] = jsonData.message === \"No external properties file has been uploaded. Cannot provide a download at this time.\";"
                                                                ]
                                                        }
                                                }
                                        ],

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6256,96 +6257,112 @@
...
+								"key": "Content-Type",
+								"value": "application/json",
+								"description": ""
+							}
+						],
+						"body": {
+							"mode": "raw",
+							"raw": "{\"jvmId\":{{JvmId}},\"systemProperties\":\"\",\"jvmName\":\"{{JvmName}}\",\"hostName\":\"{{host}}\",\"statusPath\":\"/tomcat-power.gif\",\"httpPort\":\"9130\",\"httpsPort\":\"9131\",\"redirectPort\":\"9132\",\"shutdownPort\":\"9133\",\"ajpPort\":\"9134\",\"userName\":\"\",\"encryptedPassword\":\"\",\"jdkMediaId\":{{JdkMediaId}},\"groupIds\":[{\"groupId\":{{GroupId}}}]}"
 						},
 						"description": ""
```
jmahmoodcc: Please add a description of why this REST API needs to be called, I'm not understanding the context here - is it setup for upgrading a JVM's JDK?

The values used in the raw text of the body don't appear to be setting up a JDK test.

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6256,96 +6257,112 @@
...
-									"key": "type",
-									"value": "TOMCAT",
-									"type": "text"
-								},
-								{
-									"key": "mediaArchiveFile",
+									"key": "metaData",
 									"type": "file",
-									"src": "{{tomcatMediaSrc}}",
+									"src": "../resources/postman/setenv_upgradeJDK.bat.json",
```
jmahmoodcc: We still have an issue here as this meta data file changes the setenv.bat for all of the JVMs in the Postman-Test-Group-update group.

This test only needs to be run on a single JVM, so the changes should be localized to a single JVM and not all of the JVMs in the group. Because of all of the dependencies that exist within this test suite, we need to make sure that any changes to resources that are being reused are done within the smallest scope possible.

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6358,21 +6375,23 @@
 						}
 					],
 					"request": {
-						"url": "{{url}}/v1.0/jvms/Postman-test-jvm2/jdk",
+						"url": "{{url}}/v1.0/jvms/{{JvmName2}}/jdk",
```
jmahmoodcc: I still don't see where the JvmName2 JVM has its JDK media changed to a different version before doing the upgrade. This test still looks like it's running the scenario where the JVM's JDK version hadn't changed and the upgrade JDK API was called - which is a valid test, but we're still not covering the core feature scenario.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -727,9 +727,9 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
 	/**
 	 * This method upgrades the JDK version for the mentioned JVM
 	 * 
-	 * @param jvmName
-	 * @param user
-	 * @return Jvm
+	 * @param jvmName - name of the JVM that needs to be upgraded
```
jmahmoodcc: The hyphen between the parameter name and the description is not needed. Look at the other javadoc comments in this class for examples of correct formatting.

jwala-services/src/main/java/com/cerner/jwala/service/jvm/impl/JvmServiceImpl.java
```diff
@@ -727,9 +727,9 @@ protected void deployScriptsToUserJwalaScriptsDir(Jvm jvm, User user) throws Com
 	/**
 	 * This method upgrades the JDK version for the mentioned JVM
```
jmahmoodcc: Add more to this description here stating that the JVM is expected to be in a STOPPED state, and that the JVM is expected to be configured with a setenv.bat/sh file in order to update the JAVA_HOME.

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
```diff
@@ -1198,6 +1198,13 @@ public void testUpgradeJDK() throws CommandFailureException, IOException {
 		when(Config.mockResourceService.getResourceContent(resourceIdentifier)).thenReturn(resourceContent);
 		Jvm response = jvmService.upgradeJDK(mockJvm.getJvmName(), Config.mockUser);
 		assertEquals(response.getJvmName(), mockJvm.getJvmName());
+
+		/**
```
jmahmoodcc: Remove this comment.

Some of the the verification steps you need to assert are actually detailed in the comment:
 - resource service fetches the setenv.bat
 - resource service generates the setenv.bat

Please add those verify statements for the mockResourceService.

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
```diff
@@ -1198,6 +1198,13 @@ public void testUpgradeJDK() throws CommandFailureException, IOException {
...
+		 * The persistence Service will be called three times - to fetch JVM details by
+		 * NAME-To fetch Setenv.bat Resource content - To generate the resource file-
+		 * Setenv.bat . JDK upgrade is assumed to be successful,if all these three steps
+		 * happens without error
+		 **/
 		verify(Config.mockJvmPersistenceService, times(3)).findJvmByExactName(anyString());
```
jmahmoodcc, line 1202: There are other verify statements missing.

When you go through the upgradeJDK method there are interactions with other services (that are correctly mocked for this test case) that need to be verified. This is making sure that we have automated code coverage for the core feature at the JvmService level.

- the JvmControlService can be verified to make sure it checks for a STOPPED state, and that the Windows/Linux service is deleted and then created again later
 - the BinaryDistributionService can be verified to make sure that it is distributing the JDK media

The test doesn't need to physically perform these steps (i.e. actually pushing files to the file system), but it does need to verify that the mocked helper services that would perform those functions are being called as expected.

jmahmoodcc (changes requested):
We need to make sure our automated tests are adding value by covering core functionality. 

We also need to make sure that we are making our tests modular and have as few dependencies as possible on other tests. In the past we did not do this and we're paying for it now. Going forward let's not repeat those same mistakes.

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
Karthik263: I have refactored and create a group and JVM specifically for JVM upgrade now as per your suggestions.

jwala-services/src/test/java/com/cerner/jwala/service/jvm/impl/JvmServiceImplTest.java
Karthik263: These changes were addressed

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
Karthik263: I have refactored the Postman test cases for JDK upgrade to be modular. I am creating a a new group and a JVM under the group solely for JDK upgrade.Once the test execution is done, I have added tear down operation for the created JVM and group. These set of operation do not depend upon other existing Groups/JVM and can be run independently as per your suggestion.

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6256,47 +6259,172 @@
...
+						"header": [
+							{
+								"key": "Content-Type",
+								"value": "application/json",
+								"description": ""
+							}
+						],
+						"body": {
+							"mode": "raw",
+							"raw": "{\"jvmName\":\"{{JDKUpgradeJvmName}}\",\"groupIds\":[{\"id\":10,\"groupId\":{{JDKUpgradeGroupId}}}],\"hostName\":\"{{host}}\",\"statusPath\":\"/tomcat-power.gif\",\"systemProperties\":\"\",\"httpPort\":\"10130\",\"httpsPort\":\"10131\",\"redirectPort\":\"10132\",\"shutdownPort\":\"10133\",\"ajpPort\":\"10134\",\"userName\":\"\",\"encryptedPassword\":\"\",\"jdkMediaId\":{{JdkMediaId}},\"tomcatMediaId\":{{TomcatMediaId}}}"
```
jmahmoodcc: Don't hardcode the ports.
When all the JVM properties are set as variables it's much easier to find duplicates or update them in the environment file than searching through this collection.

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6256,47 +6259,172 @@
...
+							"listen": "test",
+							"script": {
+								"type": "text/javascript",
+								"exec": [
+									"tests[\"Status code is 500\"] = responseCode.code === 500;",
+											"",
+									"var jsonData = JSON.parse(responseBody);",
+									"tests[\"Assert 500 error\"] = jsonData.status === 500;",
+
+									"tests[\"Assert No setenv.bat error\"] = jsonData.message === \"The remote JVM ::Postman-test-jvm-jdk-upgrade must be in STOPPED state before attempting to upgrade the JDK of the JVM\";"
```
jmahmoodcc: Use the environment variable for the JVM name here

jmahmoodcc: Also we need to fix the formatting of the message. Don't use 2 colons here, you can just use the JVM name. This will be a fix here and in the Java code base.

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6256,47 +6259,172 @@
...
+							"listen": "test",
+							"script": {
+								"type": "text/javascript",
+								"exec": [
+									"tests[\"Status code is 500\"] = responseCode.code === 500;",
+											"",
+									"var jsonData = JSON.parse(responseBody);",
+									"tests[\"Assert 500 error\"] = jsonData.status === 500;",
+
+									"tests[\"Assert No setenv.bat error\"] = jsonData.message === \"setenv.bat resource file does not exist for JVM: Postman-test-jvm-jdk-upgrade ,and the file is required to set the JAVA_HOME for the JVM.\";"
```
jmahmoodcc: Again, don't hard code the JVM name here.

jmahmoodcc: Also change the formatting of the message:
"The setenv.bat file does not exist as a resource for JVM: Postman-test-jvm-jdk-upgrade. Please create a setenv.bat resource for the JVM in order to set the JAVA_HOME environment variable."

jmahmoodcc: Will this fail for Linux installs since the JVMs will be configured with a setenv.sh?

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6319,33 +6447,104 @@
 							"formdata": [
 								{
 									"key": "name",
-									"value": "Postman-Test-JDK-Media-update",
+									"value": "{{JdkMediaName}}2",
```
jmahmoodcc: Don't create variable names that are appended with numbers, and definitely don't reference them in this manner where a number is being appended on the end.
Use variable names that are meaningful and provide context. Modern compilers are smart and will do performant variable name replacement so we don't have to.

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6319,33 +6447,104 @@
...
+						"header": [
+							{
+								"key": "Content-Type",
+								"value": "application/json",
+								"description": ""
+							}
+						],
+						"body": {
+							"mode": "raw",
+							"raw": "{\"jvmId\":{{JDKUpgradeJvmId}},\"systemProperties\":\"\",\"jvmName\":\"{{JDKUpgradeJvmName}}\",\"hostName\":\"{{host}}\",\"statusPath\":\"/tomcat-power.gif\",\"httpPort\":\"10131\",\"httpsPort\":\"10132\",\"redirectPort\":\"10133\",\"shutdownPort\":\"10134\",\"ajpPort\":\"10135\",\"userName\":\"\",\"encryptedPassword\":\"\",\"jdkMediaId\":{{JdkMediaId2}},\"tomcatMediaId\":{{TomcatMediaId}},\"groupIds\":[{\"groupId\":{{JDKUpgradeGroupId}}}]}"
```
jmahmoodcc: Fix hard coded ports

jmahmoodcc (changes requested):
Just some minor changes needed in the postman tests, but overall looks much better. 
Nice work!

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6256,6 +6262,250 @@
...
+						"header": [
+							{
+								"key": "Content-Type",
+								"value": "application/json",
+								"description": ""
+							}
+						],
+						"body": {
+							"mode": "raw",
+							"raw": "{\"jvmName\":\"{{JDKUpgradeJvmName}}\",\"groupIds\":[{\"id\":10,\"groupId\":{{JDKUpgradeGroupId}}}],\"hostName\":\"{{host}}\",\"statusPath\":\"/tomcat-power.gif\",\"systemProperties\":\"\",\"httpPort\":\"{{jdkUpgrade-HTTP-port}}\",\"httpsPort\":\"{{jdkUpgrade-HTTPS-port}}\",\"redirectPort\":\"{{jdkUpgrade-redirect-port}}\",\"shutdownPort\":\"{{jdkUpgrade-shutdown-port}}\",\"ajpPort\":\"{{jdkUpgrade-ajp-port}}\",\"userName\":\"\",\"encryptedPassword\":\"\",\"jdkMediaId\":{{JdkMediaId}},\"tomcatMediaId\":{{TomcatMediaId}}}"
```
jmahmoodcc: Please follow the camel case formatting for variable names.
Don't use hyphens in the variable names for the ports.

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6256,6 +6262,250 @@
...
+							"listen": "test",
+							"script": {
+								"type": "text/javascript",
+								"exec": [
+									"tests[\"Status code is 500\"] = responseCode.code === 500;",
+											"",
+									"var jsonData = JSON.parse(responseBody);",
+									"tests[\"Assert 500 error\"] = jsonData.status === 500;",
+
+									"tests[\"Assert No setenv.bat error\"] = jsonData.message === \"The setenv.bat file does not exist as a resource for JVM: \"+pm.variables.get(\"JDKUpgradeJvmName\")+\". Please create a setenv.bat resource for the JVM in order to set the JAVA_HOME environment variable.\"|| jsonData.message === \"The setenv.sh file does not exist as a resource for JVM: \"+pm.variables.get(\"JDKUpgradeJvmName\")+\". Please create a setenv.sh resource for the JVM in order to set the JAVA_HOME environment variable.\";"
```
jmahmoodcc: This line is way too long.
Break up each assertion message into its own variable: one for windows and one for linux, and then created the conditional check from those variables.

jmahmoodcc: Use the targetOs variable to determine if the test is for windows or linux:
https://github.com/cerner/jwala/blob/master/jwala-integration-test/src/test/postman/hostUserPassword_Environment.js#L10

https://github.com/cerner/jwala/blob/master/jwala-integration-test/src/test/postman/hostUserPassword_Environment.js#L95

https://github.com/cerner/jwala/blob/master/jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json#L1037

jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
```diff
@@ -6256,6 +6262,250 @@
...
+						"header": [
+							{
+								"key": "Content-Type",
+								"value": "application/json",
+								"description": ""
+							}
+						],
+						"body": {
+							"mode": "raw",
+							"raw": "{\"jvmId\":{{JDKUpgradeJvmId}},\"systemProperties\":\"\",\"jvmName\":\"{{JDKUpgradeJvmName}}\",\"hostName\":\"{{host}}\",\"statusPath\":\"/tomcat-power.gif\",\"httpPort\":\"{{jdkUpgrade-HTTP-port}}\",\"httpsPort\":\"{{jdkUpgrade-HTTPS-port}}\",\"redirectPort\":\"{{jdkUpgrade-redirect-port}}\",\"shutdownPort\":\"{{jdkUpgrade-shutdown-port}}\",\"ajpPort\":\"{{jdkUpgrade-ajp-port}}\",\"userName\":\"\",\"encryptedPassword\":\"\",\"jdkMediaId\":{{JdkUpgradeMediaId}},\"tomcatMediaId\":{{TomcatMediaId}},\"groupIds\":[{\"groupId\":{{JDKUpgradeGroupId}}}]}"
```
jmahmoodcc: See previous comment about formatting variable names for port

jwala-integration-test/src/test/postman/jwala.postman_environment.json
```diff
@@ -139,6 +139,56 @@
       "key": "WebserverHttpsPort",
       "value": "446",
       "type": "text"
+    },
+       {
+      "enabled": true,
+      "key": "jdkUpgrade-HTTP-port",
```
jmahmoodcc: Use camel case for variable name

jwala-integration-test/src/test/postman/jwala.postman_environment.json
```diff
@@ -139,6 +139,56 @@
...
+      "value": "10130",
+      "type": "text"
+    },
+     {
+      "enabled": true,
+      "key": "jdkUpgrade-HTTPS-port",
```
jmahmoodcc: Use camel case for variable name

jwala-integration-test/src/test/postman/jwala.postman_environment.json
```diff
@@ -139,6 +139,56 @@
...
+      "value": "10131",
+      "type": "text"
+    },
+    {
+      "enabled": true,
+      "key": "jdkUpgrade-redirect-port",
```
jmahmoodcc: Use camel case for variable name

jwala-integration-test/src/test/postman/jwala.postman_environment.json
```diff
@@ -139,6 +139,56 @@
...
+      "value": "10132",
+      "type": "text"
+    },
+     {
+      "enabled": true,
+      "key": "jdkUpgrade-shutdown-port",
```
jmahmoodcc: Use camel case for variable name

jwala-integration-test/src/test/postman/jwala.postman_environment.json
```diff
@@ -139,6 +139,56 @@
...
+      "value": "10133",
+      "type": "text"
+    },
+      {
+      "enabled": true,
+      "key": "jdkUpgrade-ajp-port",
```
jmahmoodcc: Use camel case for variable name

jmahmoodcc (changes requested):
Some minor changes needed. Please see the inline comments for details.

jmahmoodcc (changes requested):
Sorry, I wasn't clear about the last set of changes I wanted you to make.

Instead of uploading both setenv.bat and setenv.sh, I was hoping you would have followed the example of the httpdconf.tpl in combination with the targetOs variable, and only upload the template that is needed for that OS. This will require you to move the setenv.sh and setenv.bat into the linux and windows directories under test/resources/postman.

Look at the github links I posted in the last review to see how its done for httpdconf.tpl.

jmahmoodcc (approved):

rahulsayini (approved):

PR #211, opened 2018-11-29
Update Selenium test case to view httpd.conf file 

rahulsayini:
Update Selenium test case to view httpd.conf file


rahulsayini:
It is decided not to do too many retries for selenium test cases

PR #212, opened 2018-11-30
Fix selenium test httpd conf

rahulsayini:
The CTP CI’s selenium test cases are breaking because it is unable to see httpd.conf page after onclick httpd.conf this is intermittent and we are assuming the selenium IE web drivers is causing it.

### DIFF
--- a/jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/operations/controlWebServer.feature
+++ b/jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/operations/controlWebServer.feature
@@ -64,6 +64,7 @@ Scenario: Do a start, status, view httpd.conf, stop and deletion of a web server
 
     # test view httpd.conf
     When I click the "httpd.conf" link of web server "CONTROL-WEBSERVER-TEST-W" under group "CONTROL-WEBSERVER-TEST-G" in the operations tab
+    @ignore
     Then I see the httpd.conf
 
     When I click the drain button of "CONTROL-WEBSERVER-TEST-W" webserver under "CONTROL-WEBSERVER-TEST-G" group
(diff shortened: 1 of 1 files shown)


jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/operations/controlWebServer.feature
```diff
@@ -64,7 +64,8 @@ Scenario: Do a start, status, view httpd.conf, stop and deletion of a web server
 
     # test view httpd.conf
```
jmahmoodcc: I would delete all 3 lines (includes the comment) - if we're not validating the result then don't even bother running the test.

BUT, before doing that: do we have a postman test that is calling the same API and verifying the response body?

jmahmoodcc (changes requested):
I would argue that we should just remove the whole test.
Do we know if any current application teams are scripting against this functionality?
Or is it only used by a manual operator?

Also, here's a good read on handling automated tests that are flaky. It's from Martin Fowler, he knows a thing or two about programming:
https://martinfowler.com/articles/nonDeterminism.html

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/operations/controlWebServer.feature
rahulsayini: Okay, I will remove the test case and we have postman test case  which will assert for Http response https://github.com/cerner/jwala/blob/8a748241b45a5565714eaacb0d532913eaaa2353/jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json#L2017

tmomalar (approved):
looks good. We can add these tests later, for now we will test manually

ManickPrakash (approved):
Looks good, We can add these tests later, for now, we will test it manually

jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/operations/controlWebServer.feature
jmahmoodcc: That's the test for the REST API to get the content - that's a different call than the Selenium test was failing on.
Here's the actual test in postman: https://github.com/cerner/jwala/blob/master/jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json#L3103

The point being: we are covered in postman.

jmahmoodcc (approved):

PR #215, opened 2019-03-28
Removed deprecated states for components that implement Lifecycle in JvmStateReceiverAdapter

Debdanr:
The Life cycle states 
MUST_DESTROY and MUST_STOP have been deprecated in Tomcat 9. Hence, removed the declarations.
So, that Jwala works with Tomcat 9.0.14.


jmahmoodcc (commented):
Looks good - just one question: were there no unit tests for this class?

jmahmoodcc:
Since the mapping of the states to the JvmStates is an abstraction on the Tomcat 7 LifecycleStates there appears to be no need for individual Tomcat states as long as the JvmStates are covered

jmahmoodcc (approved):

PR #216, opened 2019-04-16
Fix postman test

jmahmoodcc:
The postman tests were failing because the web server was sending back an unknown status in response to a stop command.

There were 2 solutions that worked:
1. Remove the test
2. Add more delay between the tests

In looking at the series of steps leading up to the sending of the stop command, I couldn't really understand what the tests were trying to accomplish - hence Option 1. But, I assumed the test was exercising some sort of edge case and I didn't want to lose the code coverage.

Option 2 is slightly better in that we don't lose the code coverage, but now we're dealing with a fragile test. Could this break again in the future? Maybe. If so, then we really need to consider doing a proper clean up.

### DIFF
--- a/jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
+++ b/jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
@@ -5603,6 +5603,17 @@
 				},
 				{
 					"name": "error: generate and deploy webserver when started exception",
+					"event": [
+						{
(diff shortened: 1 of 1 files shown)


Karthik263 (approved):
I think adding 5 seconds delay in the tests should be ok for now. We will need to analyze more why there is a trouble in starting server groups ,if there are more JVMs /servers in a group.

PR #217, opened 2019-04-23
Fix the postman tests by starting the group first

jmahmoodcc:
First, revert the changes from the last commit.

Then, start the test group before running the stop tests. 

I'm not sure if this was a mistake from the initial creation of the tests, but at least now the sequence of start and stops make sense. How did this ever work before???

### DIFF
--- a/jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
+++ b/jwala-integration-test/src/test/postman/jwala-collection.postman_collection.json
@@ -5616,39 +5616,6 @@
 					"response": []
 				},
 				{
-					"name": "stop webserver",
-					"event": [
-						{
-							"listen": "test",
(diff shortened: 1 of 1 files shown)


Karthik263 (commented):
I am able to see only the start operation over here and unable to see the removal of delay in the webserver start ..

jmahmoodcc:
@Karthik263 Here is the removal of the delay from the previous commit:
https://github.com/cerner/jwala/pull/217/files#diff-e637f1e327eceb5e7ed6106aae4dc94eL5612

Karthik263:
Okie.. For some reasons ,the changes were not getting reflected in the consolidated "Files changed" view. Now ,I am able to spot it .!

Karthik263 (approved):

PR #224, opened 2019-11-22
Updated onWebAppNameClick to read secure formData from correct json node

dskcode:
The Edit Web Apps dialog is not displaying the Secure value correctly.
The value in the database is set correctly, but on retrieval is being displayed incorrectly.


jmahmoodcc:
Good find!
Can you add a test Selenium test as well which would have prevented this to begin with: https://github.com/cerner/jwala/blob/master/jwala-integration-test/src/test/resources/com/cerner/jwala/ui/selenium/configuration/manageWebApp.feature

dskcode:
I am working on the selenium test, have to do env setup for the same on one of the spare 2016 server.

dskcode:
@jmahmoodcc I have added selenium test cases please review the same and merge the changes if it looks good,

jmahmoodcc (approved):
Looks good!

PR #225, opened 2019-12-03
Application update throws an exception in the UI

jmahmoodcc:
On an application update, the default behavior is to also update the war file that is configured with the application. When the war hasn't been configured yet an exception is thrown.

The fix is to check for a war before doing an update.

There are also some inadvertent formatting updates in the tests, the key change is the addition of a new unit test to check for a missing war on an application update.


Karthik263 (approved):
Looks good !

dskcode (approved):
Changes look good.

